PR #2, opened 2022-05-31
Minimal indico charm

arturo-seijas:


tests/test_charm.py
```diff
@@ -17,50 +13,243 @@ def setUp(self):
...
+                        'SMTP_USE_TLS': False,
+                        'INDICO_SUPPORT_EMAIL': 'example@email.local',
+                        'INDICO_PUBLIC_SUPPORT_EMAIL': 'public@email.local',
+                        'INDICO_NO_REPLY_EMAIL': 'noreply@email.local',
+                    },
+                },
+            },
+        }
+        updated_plan = self.harness.get_container_pebble_plan("indico").to_dict()
+        self.assertEqual(expected_plan, updated_plan)
```
DomFleischmann, line 210: Instead of creating the full layer datastructure and asserting it, I would recommend you to check the specific values you are looking for, e.g.:
```python
updated_plan_environment = updated_plan["services"]["indico"]["environment"]
assert updated_plan_environment["INDICO_SUPPORT_EMAIL"] == "example@email.local"
```
This way you have more granular tests which won't break if you add something else to the layer in the future. The example might not exactly be what you want to assert here, but I just wanted to show how I usually make these assertions.

.flake8
```diff
@@ -1,9 +1,10 @@
 [flake8]
-max-line-length = 99
+max-line-length = 120
 select: E,W,F,C,N
 exclude:
   venv
   .git
   build
   dist
+  lib
```
mthaddon, line 9: This feels a bit broad - how about excluding lib/charms

.flake8
```diff
@@ -1,9 +1,10 @@
 [flake8]
-max-line-length = 99
+max-line-length = 120
```
mthaddon, line 2: Any reason for 120 rather than 99? I think lots of other charm projects use 99 although I know many IS projects use 120. Using 99 makes us more compatible with others if we happen to publish a library, for instance.

README.md
```diff
@@ -7,16 +7,20 @@ A Juju charm deploying and managing Indico on Kubernetes, configurable to use a
...
+    juju relate indico-k8s redis-k8s
+    juju relate indico-k8s postgresql-k8s:db
 
-## Relations
+The charm supports the `ingress` relation, which can be used with
+[nginx-ingress-integrator](https://charmhub.io/nginx-ingress-integrator/)
 
-TODO: Provide any relations which are provided or required by your charm
+    juju deploy nginx-ingress-integrator
+    juju relate indico-k8s:ingress nginx-ingress-integrator:ingress
```
mthaddon, line 22: I think we should just include this in the deploy instructions block above, no need to mention ingress separately from the DB/redis relations.

actions.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 arturo
+# Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 #
 # TEMPLATE-TODO: change this example to suit your needs.
```
mthaddon, line 4: Can you remove this?

config.yaml
```diff
@@ -1,14 +1,43 @@
...
+    description: Email address that is shown to users on the “Contact” page.
+    default: 'support@mydomain.local'
+  indico_support_email:
+    type: string
+    description: Email address of the technical manager of the Indico instance.
+    default: 'support-tech@mydomain.local'
+  site_url:
+    type: string
+    description: URL through which Indico is accessed by users.
+    default: 'http://indico.local'
```
mthaddon, line 20: I think we should default to whatever the application name is in juju. So in terms of this config item the default could just be '', and in code whenever this value is empty we use the juju application name. We should mention this in the description. Also, this seems to be a full URL including the scheme. We should think about how we're going to do TLS here in a production setting. We'll likely want to pull the TLS certificate from a kubernetes secret, but we'll need a config item to specify which one. Question is, what do we do if someone specifies `https://indico.local` and doesn't have the accompanying certificate?

config.yaml
```diff
@@ -1,14 +1,43 @@
...
+    type: string
+    description: The password to send if the SMTP server requires authentication.
+    default: ''
+  smtp_use_tls:
+    type: boolean
+    description: If enabled, STARTTLS will be used to use an encrypted SMTP connection.
+    default: true
+
+
+
```
mthaddon, line 43: Let's remove the blank lines at the end of the file.

metadata.yaml
```diff
@@ -1,23 +1,34 @@
-# Copyright 2022 arturo
-# See LICENSE file for licensing details.
-
-# For a complete list of supported options, see:
-# https://juju.is/docs/sdk/metadata-reference
-name: indico-operator
-display-name: |
-  TEMPLATE-TODO: fill out a display name for the Charmcraft store
+name: indico-k8s
```
mthaddon: As mentioned this should be `indico` rather than `indico-k8s`.

metadata.yaml
```diff
@@ -1,23 +1,34 @@
...
+display-name: Indico
+summary: Indico web application.
+docs: https://docs.getindico.io/en/stable/
```
mthaddon, line 4: The "docs" entry here refers to a discourse based documentation link, rather than a link to upstream documentation. See https://juju.is/docs/sdk/add-docs-to-your-charmhub-page for details.

metadata.yaml
```diff
@@ -1,23 +1,34 @@
...
+maintainers:
+  - launchpad.net/~canonical-is-sre
```
mthaddon, line 6: I think this should be canonical-is-devops rather than canonical-is-sre

src/charm.py
```diff
@@ -1,104 +1,244 @@
...
+        self.framework.observe(self.on.indico_pebble_ready, self._on_pebble_ready)
+        self.framework.observe(self.on.indico_celery_pebble_ready, self._on_pebble_ready)
+        self.framework.observe(self.on.indico_nginx_pebble_ready, self._on_pebble_ready)
+
+        self._stored.set_default(
+            db_conn_str=None,
+            db_uri=None,
+            db_ro_uris=[],
+            redis_relation={},
+            secret_key=repr(os.urandom(32)),
```
mthaddon, line 51: Won't this mean we end up with a different secret_key on each unit?

mthaddon (commented):
Have done an initial pass but haven't looked at tests yet, or src/charm.py in too much detail. Just wanted to provide this first round of feedback.

Thanks for the progress so far!

src/charm.py
```diff
@@ -1,104 +1,232 @@
...
+            db_uri=None,
+            db_ro_uris=[],
+            redis_relation={},
+            pebble_statuses={
+                'indico': False,
+                'indico-nginx': False,
+                'indico-celery': False,
             },
+            secret_key=repr(os.urandom(32)),
+        )
```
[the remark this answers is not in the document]
arturo-seijas: How about now?

.flake8
arturo-seijas: Changed

README.md
arturo-seijas: It is already in the same block

src/charm.py
arturo-seijas: Yes, horizontal scalability is planned for the future. This will be addresses then.

config.yaml
arturo-seijas: Adding certificates is not supported yet but planned in the future

mthaddon: Sure, I don't think we need to support it yet, just that we need to know that whatever config format we're supporting here will work when we come to add that.

src/charm.py
arturo-seijas: Done

tests/test_charm.py
```diff
@@ -17,50 +13,183 @@ def setUp(self):
...
+        self.assertEqual('support@mydomain.local', updated_plan_env["INDICO_PUBLIC_SUPPORT_EMAIL"])
+        self.assertEqual('noreply@mydomain.local', updated_plan_env["INDICO_NO_REPLY_EMAIL"])
+        self.assertEqual('', updated_plan_env["SMTP_SERVER"])
+        self.assertEqual(25, updated_plan_env["SMTP_PORT"])
+        self.assertEqual('', updated_plan_env["SMTP_LOGIN"])
+        self.assertEqual('', updated_plan_env["SMTP_PASSWORD"])
+        self.assertTrue(updated_plan_env["SMTP_USE_TLS"])
+
+        service = self.harness.model.unit.get_container("indico-celery")\
+            .get_service("indico-celery")
```
mthaddon, line 118: I'm not sure why you need to break this onto another line?

src/charm.py
```diff
@@ -1,104 +1,247 @@
...
-        the hook created in __init__.py for it, the corresponding test,
-        and the actions.py file.
-
-        Learn more about actions at https://juju.is/docs/sdk/actions
-        """
-        fail = event.params["fail"]
-        if fail:
-            event.fail(fail)
+
+    def _are_pebble_instances_ready(self):
```
mthaddon, line 106: Can you add a docstring?

config.yaml
```diff
@@ -1,14 +1,40 @@
...
+    description: Email address that is shown to users on the “Contact” page.
+    default: 'support@mydomain.local'
+  indico_support_email:
+    type: string
+    description: Email address of the technical manager of the Indico instance.
+    default: 'support-tech@mydomain.local'
+  site_url:
+    type: string
+    description: URL through which Indico is accessed by users.
+    default: 'http://indico.local'
```
mthaddon, line 20: Can we default to '' here, and in that case we use `http://{}'.format(self.app.name)`? This means that by default it's available on whatever juju application name the operator deployed with. We'll want a comment clarifying that in the description field here, or showing examples of what you might want to specify (e.g. that it's the full URL including the scheme).

mthaddon (commented):
Some very minor comments, then I think this is ready for merging

tests/test_charm.py
arturo-seijas: Formatting will be addressed in the next PR, when tox will be added. Can we ignore this for now? Also, that line is longer than 99 columns

src/charm.py
arturo-seijas: Done

config.yaml
arturo-seijas: Not possible. Indico expects a dot to be part of the configured URL. Otherwise it drops the connections

src/charm.py
```diff
@@ -1,104 +1,247 @@
...
+                },
+            },
+            'indico-nginx': {
+                "summary": "Indico nginx layer",
+                "description": "Indico nginx layer",
+                "services": {
+                    "indico-nginx": {
+                        "override": "replace",
+                        "summary": "Nginx service",
+                        "command": "nginx",
```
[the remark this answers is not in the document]
arturo-seijas: Done

config.yaml
mthaddon: We should file a bug about this, and reference that in the description (i.e. that the URL needs to contain a dot).

src/charm.py
```diff
@@ -1,104 +1,247 @@
...
 from ops.main import main
-from ops.model import ActiveStatus
+from ops.model import (
+    ActiveStatus,
+    MaintenanceStatus,
+    WaitingStatus,
+)
+from charms.nginx_ingress_integrator.v0.ingress import IngressRequires
+
+DATABASE_NAME = 'postgres'
```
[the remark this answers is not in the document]
mthaddon: This seems like a bug with the postgresql charm (or ops-lib-pgsql) if so. Can you paste some confirmation of this and we may need to raise a bug.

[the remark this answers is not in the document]
arturo-seijas: Indeed: https://github.com/canonical/ops-lib-pgsql/issues/24

config.yaml
arturo-seijas: My bad. It works but it detects is not a production configuration. I changed it in the config file given that I also need a http schema. It seemed like the cleaner approach to have both

PR #11, opened 2022-06-14
Update dockerfile comments to reference upstream issue requiring python3.9

mthaddon:
Link to the upstream issue referencing why indico requires python3.9. We expect them to support python3.10 in the next few months, at which point we can switch to the jammy-native python version.


arturo-seijas (approved):
LGTM

PR #20, opened 2022-06-21
Fix charm in waiting status when config changed triggered

arturo-seijas:


mthaddon (approved):
LGTM. At some point we may want to remove this and just query pebble status directly, but this is fine for now.

PR #25, opened 2022-06-23
Poll customization changes

arturo-seijas:


indico.Dockerfile
```diff
@@ -24,7 +24,9 @@ ENV DEBIAN_FRONTEND=noninteractive \
 
 RUN ["/bin/bash", "-c", "mkdir -p --mode=775 /srv/indico/{etc,tmp,log,cache,archive,custom}"]
 RUN /usr/local/bin/indico setup create-symlinks /srv/indico \
-    && /usr/local/bin/indico setup create-logging-config /etc
+    && /usr/local/bin/indico setup create-logging-config /etc \
+    && echo "* * * * * git -C /srv/indico/custom pull" | crontab -u indico - \
```
mthaddon, line 28: Will there always be something at that directory? What happens if someone leaves the `customization_sources_url` config value empty?

arturo-seijas: In that case the scheduled pull will not succeed as the `/srv/indico/custom` directory will be empty

mthaddon: Ok, fair enough. At some point this will be replaced by something else when there's a way to do this within pebble directly.

mthaddon (approved):
LGTM, thx

PR #28, opened 2022-07-07
Enable piwik plugin

arturo-seijas:
Piwik settings can't be configured using the config file for the time being, so they won't be exposed in juju: https://github.com/indico/indico-plugins/issues/182
The plugin will have to be configured manually through the UI


mthaddon (approved):
LGTM

PR #34, opened 2022-07-12
Add SSO support

arturo-seijas:


src/charm.py
```diff
@@ -242,12 +245,62 @@ def _get_indico_env_config(self):
...
+                }
+            }
+            env_config["INDICO_IDENTITY_PROVIDERS"] = str(identity_providers)
         env_config["INDICO_EXTRA_PLUGINS"] = ",".join(indico_plugins)
         return env_config
 
+    def _is_config_valid(self):
+        return (
+            not self.config["saml_target_url"] or UBUNTU_SAML_URL == self.config["saml_target_url"]
+        )
```
mthaddon, line 291: This function currently only checks if the saml_target_url is valid. I think we should either change this to be specifically about the saml config (`is_saml_target_url_valid`) or update it to be extendable to other config options in the future (likely by passing the message to block on).

src/charm.py
```diff
@@ -14,12 +14,13 @@
...
 from ops.main import main
-from ops.model import ActiveStatus, MaintenanceStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 from ops.pebble import ExecError
 
 DATABASE_NAME = "indico"
-PORT = 8080
 INDICO_CUSTOMIZATION_DIR = "/srv/indico/custom"
+PORT = 8080
+UBUNTU_SAML_URL = "https://login.ubuntu.com/saml/"
```
mthaddon, line 23: This is a simple string variable currently, but if we want to support other providers in the future we'd need different variables for each.

src/charm.py
```diff
@@ -242,12 +245,62 @@ def _get_indico_env_config(self):
...
+                    "trusted_email": True,
+                    "mapping": {
+                        "user_name": "username",
+                        "first_name": "fullname",
+                        "last_name": "",
+                        "email": "email",
+                    },
+                    "identifier_field": "openid",
+                }
+            }
```
mthaddon, line 283: How would these data structures support other providers in the future? Would we set different `auth_providers` and `identity_providers` variables based on which provider(s) the user wanted to support? Do we expect to only ever support one provider, or to support multiple providers?

mthaddon (changes requested):
Some comments inline. I think there's a balance here between making things too complicated now to support a theoretical future where we support multiple providers, and allowing us to add those in the future without needing to make too invasive changes.

One other question - will this _force_ logins via the configured provider, or will users still be able to create local accounts (we'd very much prefer it to be forced)?

src/charm.py
arturo-seijas: Yes, but I think is good enough for now. YAGNI maybe?

src/charm.py
arturo-seijas: Changed

arturo-seijas:
> Some comments inline. I think there's a balance here between making things too complicated now to support a theoretical future where we support multiple providers, and allowing us to add those in the future without needing to make too invasive changes.
> 
> One other question - will this _force_ logins via the configured provider, or will users still be able to create local accounts (we'd very much prefer it to be forced)?

An account will be created inside Indico and will be linked to the auth provider. Unfortunately, Indicoo accounts require both name and surname and our SSO service is storing that as a single value, so a little manual set up of the Indico account will be required. When logged in, a prepopulated form will appear to the user to complete the Indico account creation and once that is set up, the existence of this local account will be transparent to the user

mthaddon (approved):
LGTM

PR #35, opened 2022-07-13
Add support for installing external plugins

arturo-seijas:


config.yaml
```diff
@@ -10,6 +10,10 @@ options:
     type: string
     description: URL for the git repository containing the customization files
     default: ''
+  external_plugins:
+    type: string
+    description: Comma separated list of resources to be installed as in git+https://github.com/indico/indico-plugins-cern.git/#subdirectory=themes_cern. The format of each item must  be supported by pip
```
mthaddon, line 15: "resources" has special meaning in Juju, so I think we should use another word here (likely external plugins). By "as in" do you mean "e.g."? You've got a spurious extra space between "must" and "be". Finally you're missing a trailing ".". Maybe something like `Comma-separated list of external plugins to be installed, e.g. git+https://github.com/indico/indico-plugins-cern.git/#subdirectory=themes_cern. The format of each item must be supported by pip.`?

src/charm.py
```diff
@@ -277,6 +288,18 @@ def _get_current_customization_url(self):
...
+        """Install the external plugins."""
+        if plugins:
+            indico_container = self.unit.get_container("indico")
+            process = indico_container.exec(["python3.9", "-m", "pip", "install"] + plugins)
+            process.wait_output()
+            process = indico_container.exec(["indico", "setup", "list-plugins"])
+            output, _ = process.wait_output()
+            # Parse output table, discarding header and footer rows and fetching first column value
+            available_plugins = [item.split("|")[1].strip() for item in output.split("\n")[3:-2]]
+            logging.error(available_plugins)
```
mthaddon, line 301: Do we really want to log this as an error?

mthaddon (changes requested):
Some comments inline. Also, there's a note about conflicts needing resolving.

config.yaml
arturo-seijas: Fixed

PR #37, opened 2022-07-15
Add horizontal scalability

arturo-seijas:


src/charm.py
```diff
@@ -34,21 +34,27 @@ def __init__(self, *args):
...
+        self.framework.observe(
+            self.on.indico_peers_relation_changed, self._on_indico_peers_relation_changed
+        )
 
         self._stored.set_default(
             db_conn_str=None,
             db_uri=None,
             db_ro_uris=[],
             redis_relation={},
             # This key would need to be shared across instances to support horizontal scalability
```
mthaddon, line 56: I think you can remove this comment now? (Although see comment below about not using `StoredState` at all for this).

src/charm.py
```diff
@@ -320,6 +326,18 @@ def _refresh_customization_changes(self, _):
             )
             process.wait_output()
 
+    def _on_leader_elected(self, _) -> None:
+        """Handle leader-elected event."""
+        peer_relation = self.model.get_relation("indico-peers")
+        if not peer_relation.data[self.app].get("secret-key"):
```
mthaddon, line 332: Are we sure the `leader-elected` event only fires on the unit that's been elected as a leader? If not, we should also check this is only going to be executed on the leader.

src/charm.py
```diff
@@ -320,6 +326,18 @@ def _refresh_customization_changes(self, _):
...
+            peer_relation.data[self.app].update({"secret-key": repr(os.urandom(32))})
+
+    def _on_indico_peers_relation_changed(self, event) -> None:
+        """Handle relation-changed event for the indico_peers relation."""
+        secret_key = event.relation.data[self.app].get("secret-key")
+        if secret_key and secret_key != self._stored.secret_key:
+            self._stored.secret_key = secret_key
```
mthaddon, line 339: I think we could actually simplify this by not using StoredState at all here, which would mean this function isn't needed at all - instead in the `env_config` we'd just do `"SECRET_KEY": self.model.get_relation("indico-peers").data[self.app].get("secret-key"),`.

mthaddon (commented):
Some comments inline

src/charm.py
arturo-seijas: Yes, my bad

PR #38, opened 2022-07-15
Migrate the database automatically

arturo-seijas:
Add `indico db upgrade` and `indico db --all-plugins upgrade` to apply on deployment both the indico and the plugin schema and data changes


mthaddon (approved):
LGTM, thx

PR #39, opened 2022-07-15
Remove tracking pebble status on stored state

arturo-seijas:


src/charm.py
```diff
@@ -247,19 +243,22 @@ def _get_indico_env_config(self):
...
+            event.defer()
+            return
+
+        self.model.unit.status = MaintenanceStatus("Configuring pod")
+        for container_name in self.model.unit.containers:
+            self._config_pebble(self.unit.get_container(container_name))
+
+        self._download_customization_changes()
+        self.ingress.update_config(self._make_ingress_config())
+        self.model.unit.status = ActiveStatus()
```
mthaddon, line 261: You're setting state to `ActiveStatus()` here even if `_config_pebble` a few lines above has set state to `WaitingStatus("Waiting for pebble")`. That seems like something we'd want to avoid?

mthaddon (changes requested):
One comment inline

src/charm.py
arturo-seijas: Not sure if I follow. There's a return statement right after `WaitingStatus("Waiting for pebble")`

mthaddon: Not in the `_config_pebble` function I think?

arturo-seijas: Yes, at this point this is the easiest way to track it since `_config_pebble`is also called when `on_pebble_ready` is triggered. With the indico and celery configs diverging it makes sense to evaluate a refactoring, but I feel this is out of the scope of this PR

mthaddon: Ok, agreed this is fine for now.

mthaddon (approved):
LGTM

PR #42, opened 2022-07-21
Remove check-libs step from pull request workflow

arturo-seijas:
The check-libs step requires a charmhub token to be configured in every forked repository. To avoid the need of manual set up, this step is removed from the pull request workflow.


mthaddon (approved):
LGTM, thx

PR #49, opened 2022-07-28
Add manual workflow to publish feature branches

arturo-seijas:
- Add manual step to publish a feature branch in a track in latest/edge/
- Build image is moved to a github action
- Publish resource image is moved to a github action

Unfortunately, github does not currently allow more nested workflow levels so there's some duplication in both publish workflows.


.github/workflows/publish_branch.yaml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Upload charm to charmhub
+        uses: canonical/charming-actions/upload-charm@1.0.3
+        with:
+          credentials: ${{ secrets.charmhub-token }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          channel: ${{ needs.select-branch.outputs.branch }}
+          upload-image: false }}
```
mthaddon, line 73: `false }}` looks a bit odd, is that intended?

mthaddon (commented):
One comment inline, but I also don't quite understand why we're deleting (for instance) on_pull_request.yaml and publish.yaml. Would be good to have a more thorough description of the changes in the description field of the PR.

.github/workflows/publish_branch.yaml
arturo-seijas: Good catch. I can't test the manual step until merged.

arturo-seijas:
> One comment inline, but I also don't quite understand why we're deleting (for instance) on_pull_request.yaml and publish.yaml. Would be good to have a more thorough description of the changes in the description field of the PR.

The reason is that the trigger is defined now in the ci.yaml file

mthaddon (approved):
LGTM, thx

mthaddon (approved):
LGTM

PR #50, opened 2022-07-29
Add grafana integration (COS)

arturo-seijas:


mthaddon:
Do we not need to actually include a dashboard of some sort?

arturo-seijas:
> Do we not need to actually include a dashboard of some sort?

Not necessarily. We could add it later.

metadata.yaml
```diff
@@ -49,4 +51,4 @@ requires:
 
 peers:
   indico-peers:
-    interface: indico-instance
+    interface: indico-instance
```
jdkandersson, line 54: Do we need this dangling whitespace? ;)

jdkandersson (commented):
I'm guessing the .gitkeep is intended to ensure that the `grafana_dashboards` folder stays? Is it required to be there when the charm is running?

metadata.yaml
arturo-seijas: We don't. Fixed

arturo-seijas:
> I'm guessing the .gitkeep is intended to ensure that the `grafana_dashboards` folder stays?
Yes
> Is it required to be there when the charm is running?
No

A dashboard has been added so there's no need for that file anymore.

arturo-seijas:
Sample screenshot
image

gtrkiller (dismissed):
LGTM

src/grafana_dashboards/indico.json
```diff
@@ -0,0 +1,586 @@
...
+    "iteration": 1562682051068,
+    "links": [
+        {
+            "icon": "doc",
+            "tags": [],
+            "targetBlank": true,
+            "title": "Docs",
+            "tooltip": "Official documentation of Indico",
+            "type": "link",
+            "url": "https://docs.getindico.io/en/latest/"
```
mthaddon, line 65: If this is the [Docs] link in the screenshot you're showing there, I think it may be better to have this point to the charmhub documentation page - https://charmhub.io/indico - as the stats we're looking at here are for nginx rather than indico itself, so it'll be easier for us to explain what people are seeing there.

src/grafana_dashboards/indico.json
```diff
@@ -0,0 +1,586 @@
...
+          },
+          {
+            "icon": "info",
+            "tags": [],
+            "targetBlank": true,
+            "title": "GitHub",
+            "tooltip": "Indico sources on GitHub",
+            "type": "link",
+            "url": "https://github.com/indico/indico"
```
mthaddon, line 74: I'm not sure what standard practice is here, whether we should link to https://github.com/canonica/indico-operator or the indico upstream code itself. My guess is for an operational dashboard like this it might be better to point to the charm code though. Obviously we'd want to update the tooltip as well. Maybe we list all four (Charm Docs, Indico Docs, Charm Code, Indico Code)?

src/grafana_dashboards/indico.json
arturo-seijas: Done

mthaddon (approved):
This looks great. Thanks for also including the screenshot. Will be nice to see this in action.

gtrkiller (approved):
LGTM

PR #53, opened 2022-08-04
add high level development instructions

jdkandersson:


jdkandersson:
Just found some more documentation I want to add, please hold review.

jdkandersson:
Alright, good for review now

mthaddon:
I think this is already covered by https://charmhub.io/indico/docs/contributing - can you take a look and confirm? The nature of documentation like this is still something we're in discussions with the documentation team about, but our current understanding is that this kind of content should be covered in the discourse-based detailed documentation that's rendered on charmhub.

jdkandersson:
> I think this is already covered by https://charmhub.io/indico/docs/contributing - can you take a look and confirm? The nature of documentation like this is still something we're in discussions with the documentation team about, but our current understanding is that this kind of content should be covered in the discourse-based detailed documentation that's rendered on charmhub.

Ah I see, do you think it is worth linking to the charmhub developer docs from the README file? The `juju deploy ./indico-operator_ubuntu-20.04-amd64.charm` command on charmhub is missing the resources (which it says in the comment above it as well), how do updates to that documentation work? I'm happy to add that part. The charmhub docs also seem to be missing the instructions for building the Docker images and how to get them into microk8s from Docker

.dockerignore
```diff
@@ -0,0 +1,5 @@
+# Ignore everything
```
arturo-seijas, line 1: I think It'd be a good idea to split this dockerignore changes to another PR. I don't think they match the PR scope and they affect the build process.

arturo-seijas (changes requested):

mthaddon:
> Ah I see, do you think it is worth linking to the charmhub developer docs from the README file? The `juju deploy ./indico-operator_ubuntu-20.04-amd64.charm` command on charmhub is missing the resources (which it says in the comment above it as well), how do updates to that documentation work? I'm happy to add that part. The charmhub docs also seem to be missing the instructions for building the Docker images and how to get them into microk8s from Docker

The readme does already link to that - `For further details, [see here](https://charmhub.io/indico/docs)`. I can explain how to edit the discourse docs when we catch up tomorrow.

.dockerignore
mthaddon: I agree on this.

mthaddon:
Closing out this pull request as discussed with David - changes will be integrated into the discourse-based documentation.

PR #54, opened 2022-08-05
Add bootstrap page load integration test

jdkandersson:
- Fixes issue with the lint command where pflake8 is not compatible with flak8 version 5
- Moves the building of the charm, deploying the dependencies and creating the relations into a fixture
- Adds a tests that checks whether the indico bootstrap page loads

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -35,7 +35,6 @@ async def test_active(ops_test: OpsTest, app_name: str, indico_charm):
 
     Assume that the charm has already been built and is running.
     """
-    # build and deploy charm from local source folder
     assert ops_test.model.applications[app_name].units[0].workload_status == ActiveStatus.name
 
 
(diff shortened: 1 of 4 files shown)


arturo-seijas:
LGTM, thx

arturo-seijas (dismissed):

tests/integration/test_charm.py
```diff
@@ -27,42 +28,48 @@ async def juju_run(unit, cmd):
...
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy(ops_test: OpsTest, indico_image, indico_nginx_image):
-    """Build the charm-under-test and deploy it together with related charms.
+async def test_active(ops_test: OpsTest, app_name: str, indico_charm):
+    """Check that the charm is active.
 
-    Assert on the unit status before any relations/configurations take place.
+    Assume that the charm has already been built and is running.
     """
     # build and deploy charm from local source folder
```
mthaddon, line 38: I think this comment is no longer relevant?

tests/integration/test_charm.py
```diff
@@ -27,42 +28,48 @@ async def juju_run(unit, cmd):
...
+    """Check that the bootstrap page is reachable.
+
+    Assume that the charm has already been built and is running.
+    """
+    # Read the IP address of indico
+    status = await ops_test.model.get_status()
+    unit = list(status.applications[app_name].units)[0]
+    address = status["applications"][app_name]["units"][unit]["address"]
+
+    # Send request to bootstrap page and set Host header to indico (which the application expects)
```
mthaddon, line 54: I think strictly speaking it's expecting the host header to match the `app_name` (which in this case is indico). Worth referencing that (and using it in the line below)?

mthaddon (commented):
Two very minor comments inline

tests/integration/test_charm.py
jdkandersson: Thanks for picking this up, done

tests/integration/test_charm.py
jdkandersson: Thanks for the tip, that makes sense! Done

PR #56, opened 2022-08-09
Add .dockerignore that only sends relevant files to the build process

jdkandersson:
Reduces the files that are sent to the build context by getting docker to ignore all files except for the `files` directory.


arturo-seijas:
LGTM

arturo-seijas (approved):

PR #60, opened 2022-08-11
Prevent GitHub actions that require Charmhub secret to run on a forked repo

weiiwang01:


.github/workflows/on_push.yaml
```diff
@@ -16,6 +16,7 @@ on:
 
 jobs:
   lib-check:
+    if: github.repository == 'canonical/indico-operator'
```
arturo-seijas, line 19: Can we disable the whole on push listener instead?

arturo-seijas (changes requested):

.github/workflows/on_push.yaml
mthaddon, line 19: I think you can reference `repository_owner` here, which might make this a little more portable (i.e. we could have `canonical` and reuse for other repos without needing a change.

weiiwang01: I don't think the push event filter can do filtering based on the owner, and there if statement can't apply to the entire workflow, only to jobs. `lib-check` job is the precondition of all the jobs that use the Charmhub token, disable this one will prevent all of them from running which leaves only test jobs.

arturo-seijas: Thx for the clarification

weiiwang01: Good idea, updated, thanks

.github/workflows/on_push.yaml
```diff
@@ -16,6 +16,7 @@ on:
 
 jobs:
   lib-check:
+    if: github.repository_owner == 'canonical'
```
mthaddon, line 19: I think it'd be good to have a comment explaining either here or in the comment block towards the top of the file that we don't want charm publishing to happen on forked repos, and this will achieve this (in case we change the layout and to make it easier to understand the intent).

weiiwang01: Good idea, add some comments, thx

arturo-seijas (approved):
LGTM

mthaddon (approved):
LGTM, thx

PR #61, opened 2022-08-11
Fix failing integration test

arturo-seijas:
Adapt the integration tests to the breaking change introduced in pylibjuju, as explained in https://github.com/juju/python-libjuju/issues/707#issuecomment-1212490521


mthaddon (approved):
LGTM, thx!

PR #63, opened 2022-08-16
Pin charmcraft to version 1.x

arturo-seijas:
Pin charmcraft version to 1.x to workaround the authentication issues experienced when uploading images to charmhub with version 2.0.0


mthaddon (approved):
LGTM - as mentioned we'll want to file an issue against charmcraft if this works

PR #67, opened 2022-08-17
Add prometheus integration

arturo-seijas:


merkata:
Looks really good, the PR is very clear in terms of changes. I like how container events on pebble collapse into a single handler function, seems like a good pattern :) 
Do you plan on having an integration test for prometheus, seeing it scrapes something etc, is that planned?
Good work!

arturo-seijas:
> Looks really good, the PR is very clear in terms of changes. I like how container events on pebble collapse into a single handler function, seems like a good pattern :) Do you plan on having an integration test for prometheus, seeing it scrapes something etc, is that planned? Good work!

No, I don't plan to add an integration test for prometheus. The scraping falls under the prometheus charm responsibility hence, that'd be the proper place to add it.
Thank you for your kind words :)

mthaddon:
Seems like the integration tests failed here - can you take a look?

arturo-seijas:
> Seems like the integration tests failed here - can you take a look?

Fixed

files/etc/nginx/nginx.conf
```diff
@@ -37,5 +37,9 @@ http {
       add_header 'Content-Type' 'application/json';
       return 204;
     }
+
+    location = /stub_status {
+      stub_status;
+    }
```
mthaddon, line 43: I wonder if we need to/can restrict this to only being accessible from 127.0.0.1? I don't think we'd want that URL to be accessible to end users would we?

arturo-seijas: Indeed. Good catch

mthaddon (approved):
Looks great, thanks!

PR #68, opened 2022-08-17
Chore: specify version of python to use in CI

jdkandersson:
Explicitly specifies the Python version to use in CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,33 +7,27 @@ on:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
-        with:
-          python-version: '3.8'
       - name: Install dependencies
-        run: python -m pip install tox
+        run: python3 -m pip install tox
       - name: Run linters
         run: tox -e lint
   unit-test:
     name: Unit tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
-        with:
-          python-version: '3.8'
       - name: Install dependencies
-        run: python -m pip install tox
+        run: python3 -m pip install tox
       - name: Run tests
         run: tox -e unit
   build-indico-image:
     name: Build indico image
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     outputs:
       image-id: ${{ steps.build.outputs.imageid }}
     steps:
@@ -45,7 +39,7 @@ jobs:
           resource-name: indico-image
   build-indico-nginx-image:
     name: Build nginx image
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     outputs:
       image-id: ${{ steps.build.outputs.imageid }}
     steps:
@@ -57,7 +51,7 @@ jobs:
           resource-name: indico-nginx-image
   integration-test-microk8s:
     name: Integration tests (microk8s)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     needs: [build-indico-image, build-indico-nginx-image, lint, unit-test]
     steps:
       - name: Checkout
@@ -80,9 +74,6 @@ jobs:
           imageId=$(docker load --input indico-nginx-image/indico-nginx-image.tar | sed -e "s/^Loaded image ID: //")
           docker tag $imageId localhost:32000/indico-nginx:latest
           docker push localhost:32000/indico-nginx:latest
-      - uses: actions/setup-python@v4
-        with:
-          python-version: '3.8'
       - name: Run integration tests
         run: tox -e integration -- --model testing --indico-image localhost:32000/indico:latest --indico-nginx-image localhost:32000/indico-nginx:latest
       - name: Dump logs
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.8'
       - name: Install dependencies
         run: python -m pip install tox
       - name: Run linters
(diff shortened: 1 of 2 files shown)


jdkandersson:
I'm not sure why the lint command failed? Maybe because no Python files were changed?

jdkandersson:
Actually, the problem seems to be that codespell thinks there is a spelling mistake in the licence file:
`./LICENSE:128: complies ==> compiles`

I'll remove the licence file from being considered for spell checking. I wonder why this is the first time this happened...

merkata:
Looks good and the issue with the LICENSE file is good to have in mind for other charms :)

merkata (dismissed):
The changes are nice and clear, I like that the python version (3.8) is tied to the one that is used in 20.04, so there is consistency. Very nice, approving.

tox.ini
```diff
@@ -48,7 +48,8 @@ commands =
     # codespell {[vars]lib_path}
     codespell {toxinidir}/. --skip {toxinidir}/.git --skip {toxinidir}/.tox \
       --skip {toxinidir}/build --skip {toxinidir}/lib --skip {toxinidir}/venv \
-      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg
+      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg \
+      --skip {toxinidir}/LICENSE
```
arturo-seijas, line 52: This fails due to some changes in the "rare" dictionary. It'd be better to exclude it instead of skipping the file: [--builtin clear](https://github.com/codespell-project/actions-codespell/blob/master/README.md#parameter-builtin)

.github/workflows/ci.yaml
```diff
@@ -74,6 +80,9 @@ jobs:
           imageId=$(docker load --input indico-nginx-image/indico-nginx-image.tar | sed -e "s/^Loaded image ID: //")
           docker tag $imageId localhost:32000/indico-nginx:latest
           docker push localhost:32000/indico-nginx:latest
+      - uses: actions/setup-python@v4
```
arturo-seijas, line 83: Why downgrading python? If I'm not mistaken, ubuntu jammy ships with python 3.10. Anyway, I wouldn't manage this package independently, relying in the version shipped with ubuntu, we are just adding additional complexity.

arturo-seijas (changes requested):

arturo-seijas:
> Actually, the problem seems to be that codespell thinks there is a spelling mistake in the licence file: `./LICENSE:128: complies ==> compiles`
> 
> I'll remove the licence file from being considered for spell checking. I wonder why this is the first time this happened...

I already left a comment for this but it seems to be a problem with the "rare" dictionary.

.github/workflows/ci.yaml
jdkandersson: It is to make it explicit which version is being tested, which you will see is quite confusing in this case using the implicit approach. `ubuntu-latest` still actually means 20.04: https://github.com/actions/runner-images. The alternative could be to be specific about the ubuntu version, e.g. `ubuntu-22.04` would also work. Without picking one approach, we are reliant on GitHub's meaning of `ubuntu-latest` for which version of Ubuntu and Python is being used, which could change any time. It would also be worth matching the ubuntu version in GitHub actions to the `build-on` and `run-on` in `charmcraft.yaml`.

arturo-seijas: My bad, I tought it was jammy.

> The alternative could be to be specific about the ubuntu version, e.g. ubuntu-22.04 would also work
This is what I was referring to. This would be my preferred approach because we'd be managing the version of two components together instead of independently installing another (or the same) python version on top of ubuntu.

> It would also be worth matching the ubuntu version in GitHub actions to the build-on and run-on in charmcraft.yaml
This is already in place.

jdkandersson: I'm keen for some feedback on the options we have here:

- we could add a step installing a specific Python version (such as what is proposed in this PR) or
- we could be explicit about the ubuntu version being used (i.e., `ubuntu-22.04` instead of `ubuntu-latest`) and rely on the inbuilt system Python version that comes with a given OS. This would match the `charmcraft.yaml` configuration and also ensure that the same Python version is being used in testing and when it is running.

A further question is actually whether it is possible to have the charm code run on a different version of python than the default system version, for example when Python 3.11 is released later this year do we have to wait for Ubuntu 23.04 or even 24.04 before we can use it? Or is there a way that we can get the charm to run on a specific Python version?

To enable adoption of new language features, I'm leaning towards preferring specifying the Python version if we can define the Python version the charm runs on when it is deployed. If not, I would create an issue (wherever appropriate) to request that and recommend basing it on specific Ubuntu versions for now.

@merkata @mthaddon 

When we're finished discussing it here I'll add it to our contribution guide.

tox.ini
jdkandersson: Happy to do something different, help me understand please, is the proposal to:

- Exclude the `rare` dictionary?
- Include the `complies` work from the rare dictionary?
- Something else?

.github/workflows/ci.yaml
merkata: This is a good discussion and some good points were made. I'm leaning more towards using the ubuntu version for testing in CI that is in the `charmcraft.yaml` and used for building and running. It is versioned and more concise in my opinion. We should test against the python version that is ultimately used in the other parts of the charm lifecycle (building and running).

mthaddon: I think we should be explicit about the version rather than relying on `ubuntu-latest`, which in this case I think means switching to `ubuntu-20.04`. I _think_ that main thing here is that we should be matching the version that we're running in CI with the version that we're defining in `charmcraft.yaml` as our `build-on` and `run-on` targets. Those will define what charmcraft uses to build the charm on, and what docker image is used to deploy the charm code by juju. For example, a `kubectl describe` on a deployed indico pod has the following for the charm container:
```
Containers:
  charm:
    Container ID:
    Image:         jujusolutions/charm-base:ubuntu-20.04
    Image ID:
    Port:          <none>
    Host Port:     <none>
```

tox.ini
jdkandersson: I have added complies to the list of allowed words

arturo-seijas: I was referring to the former, but your solution seems better. Anyway, the issue has been already fixed, so we can revert the change: https://github.com/projectcontour/contour-operator/pull/533

jdkandersson: Done

arturo-seijas (approved):
LGTM

merkata (commented):
Looks great!

PR #71, opened 2022-08-23
Fix external image upload

arturo-seijas:
Charmhub actions will fail if upstream-resource is present. More info in https://github.com/canonical/charming-actions/issues/60


mthaddon (dismissed):
LGTM, thx

mthaddon (approved):
LGTM

PR #72, opened 2022-08-23
Use tox default dictionaries

arturo-seijas:
Rare dictionry issues incodespell have been fixed. Tox is set up to use it again.


mthaddon (approved):
LGTM

PR #73, opened 2022-08-23
Pass proxy configuration to git and pip commands

arturo-seijas:
Add environmental proxy variables for pip and git command executions.


src/charm.py
```diff
@@ -354,6 +350,14 @@ def _get_indico_env_config(self):
             env_config["INDICO_IDENTITY_PROVIDERS"] = str(identity_providers)
         return env_config
 
+    def _get_http_proxy_configuration(self):
```
mthaddon, line 353: Can we include a docstring pls?

mthaddon (approved):
Great, thanks!

PR #74, opened 2022-08-23
Be more explicit about uses of proxy config variables

mthaddon:
This feels clearer to me as a description of where the are used, but does mean we need to update this if the uses of these config options changes in the future.


arturo-seijas (approved):
LGTM

PR #75, opened 2022-08-24
Fix celery error when enabling S3 plugin

arturo-seijas:
Fixes the error thrown due to the S3 plugin not being installed in the celery container.


mthaddon (approved):
LGTM

PR #79, opened 2022-08-24
Add workaround for proxy env config when installing git packages via pip

arturo-seijas:
Add workaround for proxy env config when installing git packages via pip. Issue https://github.com/pypa/pip/issues/11405


mthaddon (dismissed):
Looks goods, thanks

mthaddon (approved):
We may want to update this to check if the config is set and unset it if so, but this is fine for now.

PR #83, opened 2022-08-25
Create a new Github action to comment test coverage report on PR

weiiwang01:
Add a new Github action to create a comment on each pull request with the test coverage report in it. Most of the workflow code is inspired by this [post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) wroten by the Github security team.

Since only workflow completion can trigger another workflow, I created another workflow that’s only the unit tests to trigger the comment workflow instead of using the CI workflow as the trigger to shorten the feedback loop. So now we don’t need to wait till the CI workflow finishes to see the comment popping in the PR, which may take 10+ minutes. The only drawback is the unit tests are running twice, but the duration of the unit tests is about 10 seconds, which is acceptable to me.

Please check this [demo PR page](https://github.com/weiiwang01/indico-operator/pull/14) on my forked repository to see this new workflow in action.


arturo-seijas:
Nice work! I'm actually thinking if we can split the current CI workflow in integration tests + rest (linting, unit tests and coverage) and report also the linting and unit test errors. What do you think?

arturo-seijas (changes requested):
Just a remark to extend the reported results

weiiwang01:
> Nice work! I'm actually thinking if we can split the current CI workflow in integration tests + rest (linting, unit tests and coverage) and report also the linting and unit test errors. What do you think?

I was thinking about this too, but if we want to keep the dependency between unit tests and integration tests, that's only run the integration test when unit tests succeed. Things could be a little bit complicated since there's no direct way in Github action to implement dependency between workflows.

arturo-seijas:
> > Nice work! I'm actually thinking if we can split the current CI workflow in integration tests + rest (linting, unit tests and coverage) and report also the linting and unit test errors. What do you think?
> 
> I was thinking about this too, but if we want to keep the dependency between unit tests and integration tests, that's only run the integration test when unit tests succeed. Things could be a little bit complicated since there's no direct way in Github action to implement dependency between workflows.

Can't we add the linting step in the new unit_test.yaml file? We'd keep the dependency and maybe generate from there the proper artifacts to then report the status

jdkandersson:
> > > Nice work! I'm actually thinking if we can split the current CI workflow in integration tests + rest (linting, unit tests and coverage) and report also the linting and unit test errors. What do you think?
> > 
> > 
> > I was thinking about this too, but if we want to keep the dependency between unit tests and integration tests, that's only run the integration test when unit tests succeed. Things could be a little bit complicated since there's no direct way in Github action to implement dependency between workflows.
> 
> Can't we add the linting step in the new unit_test.yaml file? We'd keep the dependency and maybe generate from there the proper artifacts to then report the status

Another option could be to make it a custom GitHub action that can be run after the unit test (similar to, for example, `charmed-kubernetes/actions-operator`: https://github.com/charmed-kubernetes/actions-operator)

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run tests
+        run: tox -e unit
+      - name: Generate coverage report
+        if: always() && github.event_name == 'pull_request'
```
jdkandersson, line 19: What does always do here? A naive read makes this sounds like the equivalent in Python of `if True and github.event_name == "pull_request"` which could be abbreviated to `if github.event_name == "pull_request"`, maybe `always()` has a different meaning here?

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          mkdir -p coverage-report
+          tox -e coverage-report -q | sed -n '/___ summary ___/q;p' > ./coverage-report/coverage-report.txt
```
jdkandersson, line 22: out of curiosity, what does sed do in this case please? Is it filtering something from the coverage report that `coverage` creates?

tox.ini
```diff
@@ -65,6 +65,15 @@ commands =
         -m pytest --ignore={[vars]tst_path}integration -v --tb native -s {posargs}
     coverage report --fail-under=90
 
+[testenv:coverage-report]
+description = Create test coverage report
+deps =
+    pytest
+    coverage[toml]
```
jdkandersson, line 72: Again out of curiosity, what does the `[toml]` do here please?

.github/workflows/unit_test.yaml
weiiwang01: The always is one of the [status check functions](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) to test if the previous step succeeded. The `tox -e unit` step may fail due to test coverage being less than threshold or bug, but the test coverage report can still be generated in that case, so we want the test coverage report to be created no matter what the test result is.

.github/workflows/unit_test.yaml
weiiwang01: It was filtering some meaningless summary printed by `tox`.

tox.ini
weiiwang01: [It is needed to read `pyproject.toml`](https://coverage.readthedocs.io/en/6.4.4/config.html?highlight=coverage[toml])

.github/workflows/unit_test.yaml
jdkandersson: so what would happen if the statement was just `if: github.event_name == 'pull_request'`?

weiiwang01: If one unit test failed or if the test coverage dropped below 90%, there will be no test coverage report for you :)

jdkandersson: It turns out that, if `always()` isn't added, GitHub adds the `success()` call which is why the `always()` call is required here

.github/workflows/comment.yaml
```diff
@@ -24,29 +23,40 @@ jobs:
...
+            const coverageReport = report.reports.coverage.output.trim()
+            await createComment(`Test coverage report for ${sha}\n\`\`\`\n${coverageReport}\n\`\`\``);
+            if (!report.reports.unit.success) {
+              const unitReport = report.reports.unit.output.trim();
+              await createComment(`Unit tests failed for ${sha}\n\`\`\`${unitReport}\`\`\``);
+            }
+            if (!report.reports.lint.success) {
+              const lintReport = report.reports.lint.output.trim();
+              await createComment(`Lint checks failed for ${sha}\n\`\`\`${lintReport}\`\`\``);
+            }
```
arturo-seijas, line 62: You're missing a new line at the end here

.github/workflows/test.yaml
```diff
@@ -0,0 +1,89 @@
...
+              }
+              os.makedirs("report", exist_ok=True)
+              with open("report.json", "w+", encoding="utf-8") as o:
+                  json.dump(final_report, o, indent=2)
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: report
+          path: report.json
```
arturo-seijas, line 89: You're missing a new line at the end here

arturo-seijas (dismissed):
Looks good. Just some minor details

weiiwang01:
I added the lint check and unit test results in the comments, it will only show up if lint checks or unit tests failed.
Final results can been seen in the following four demo PR:

- [Good PR](https://github.com/weiiwang01/indico-operator/pull/51)
- [PR with failed unit tests](https://github.com/weiiwang01/indico-operator/pull/48)
- [PR with failed lint checks](https://github.com/weiiwang01/indico-operator/pull/49)
- [PR with both unit tests and lint checks failed](https://github.com/weiiwang01/indico-operator/pull/50)

Question: Can we break the dependency between unit test and integration test, that's run integration test no matter unit test success or not. In that case we don't need to run unit test twice and make the check section in PR cleaner. The only drawback is that it will be more wasteful of computing resources.

.github/workflows/comment.yaml
weiiwang01: Fixed

.github/workflows/test.yaml
weiiwang01: Fixed

arturo-seijas:
> I added the lint check and unit test results in the comments, it will only show up if lint checks or unit tests failed. Final results can been seen in the following four demo PR:
> 
> * [Good PR](https://github.com/weiiwang01/indico-operator/pull/51)
> * [PR with failed unit tests](https://github.com/weiiwang01/indico-operator/pull/48)
> * [PR with failed lint checks](https://github.com/weiiwang01/indico-operator/pull/49)
> * [PR with both unit tests and lint checks failed](https://github.com/weiiwang01/indico-operator/pull/50)
> 
> Question: Can we break the dependency between unit test and integration test, that's run integration test no matter unit test success or not. In that case we don't need to run unit test twice and make the check section in PR cleaner. The only drawback is that it will be more wasteful of computing resources.

I'm not sure I follow. I'm seeing the following workflow
image

It is my understanding that the jobs "lint" and "unit test" are already being executed as part of "unit test and lint report" and hence, duplicated, is that right? Why not removing those two?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Generate report
+        run: tox -e lint,unit,coverage-report --result-json=test-result.json
+      - name: Format test report
+        shell: python
+        env:
+          GITHUB_EVENT_PULL_REQUEST_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          GITHUB_EVENT_NUMBER: ${{ github.event.number }}
+        if: always()
+        run: |
+          import os
```
jdkandersson, line 46: Do you think this would be better as a separate Python file that is called from CI versus in-line Python code?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,89 @@
...
+    if: github.event_name == 'pull_request'
+    name: Unit test and lint report
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Generate report
+        run: tox -e lint,unit,coverage-report --result-json=test-result.json
```
jdkandersson, line 38: would it be possible to have the `lint` and `unit-test` jobs run in a way where they generate the relevant JSON reports and make them available as artefacts that can then be used by this job to format the reports and publish them? That would eliminate the need for testing and linting to run potentially twice and also simplifies this job (this step wouldn't need to install and run tox at all)

jdkandersson:
Thanks, the comments are looking excellent! Looking forward to having this available on this and our other repos

.github/workflows/test.yaml
weiiwang01: Yes, I think that's doable.

.github/workflows/test.yaml
weiiwang01: Or maybe we can use [job outputs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs) to pass test results between jobs. There's a 1MB limit but with `xz` that would be plenty for this application.

jdkandersson: Sounds like that would work also

weiiwang01:
Some additional improvements I made so far:

- Change the order of comments in PR. Now it will go with lint then unit test then test coverage
- Break the dependency between unit tests and integration tests. Now integration tests will run no matter what the unit test result is. This will help to remove duplicated unit test runs and make the checks section in PR clear.
- Use job output to pass test results between jobs, reducing the duplicated unit test runs in the "Test" workflow
- Move inline Python and JavaScript code to separate files, improving the readability and extensibility.

Test PRs with the latest workflow:

- [Good PR](https://github.com/weiiwang01/indico-operator/pull/91)
- [PR with failed lint checks](https://github.com/weiiwang01/indico-operator/pull/89)
- [PR with failed unit tests](https://github.com/weiiwang01/indico-operator/pull/88)
- [PR with both lint checks and unit tests failed](https://github.com/weiiwang01/indico-operator/pull/90)

.github/files/comment_pr.js
```diff
@@ -0,0 +1,24 @@
...
+    if (!report.reports.lint.success) {
+    const lintReport = report.reports.lint.output.trim();
+    await createComment(`Lint checks failed for ${sha}\n\`\`\`\n${lintReport}\n\`\`\``);
+    }
+    if (!report.reports.unit.success) {
+    const unitReport = report.reports.unit.output.trim();
+    await createComment(`Unit tests failed for ${sha}\n\`\`\`\n${unitReport}\n\`\`\``);
+    }
+    await createComment(`Test coverage report for ${sha}\n\`\`\`\n${coverageReport}\n\`\`\``);
+}
```
arturo-seijas, line 24: Missing line at the end of file

arturo-seijas (changes requested):

.github/files/comment_pr.js
weiiwang01: Fixed, thanks

arturo-seijas (approved):
LGTM

.github/workflows/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run tests
+        run: tox -e lint,unit,coverage-report --result-json=test-result.json
+      - name: Export test report
+        if: always() && github.event_name == 'pull_request'
+        id: export-test-report
+        run: echo "::set-output name=report::$(cat test-result.json | xz | base64 -w 0)"
+  report:
+    if: always() && github.event_name == 'pull_request'
```
mthaddon, line 25: What's the `if: always()` bit here?

weiiwang01: Since the `report` job needs the outputs from `lint-and-unit-test` job, `lint-and-unit-test` needs to be included in the `needs` section of the `report` job definition. But we still want to run `report` job even if the `lint-and-unit-test` failed, so we have to include an `always()` here or Github will insert a `success()` implicitly which will make the `report` job run only if `lint-and-unit-test` succeeded. See this document for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs

mthaddon (approved):
LGTM, thx! Really looking forward to seeing this in action.

jdkandersson (approved):

PR #85, opened 2022-08-25
Pin Indico to version 3.1

arturo-seijas:
Pin Indico version to 3.1.*


mthaddon:
Let's wait to see what the response is to https://github.com/indico/indico/issues/5471. If Indico are able to fix this before we need external users in production it feels reasonable to hold off to allow them to fix.

arturo-seijas:
Won't merge. Encoding issues are solved through #86

mthaddon (dismissed):
Thanks. Worth getting a version of this published even if we later revert back to 3.2, which we expect to do.

PR #87, opened 2022-08-25
Remove fixed docker image reference from tests

arturo-seijas:
Remove the nginx-prometheus-exporter harcoded image reference from tests.


mthaddon (approved):
LGTM

PR #88, opened 2022-08-26
Adjust pebble checks

arturo-seijas:
- Remove readiness check from prometheus exporter
- Remove readiness check from indico
- Add liveness check for indico
- Interchange readiness and liveness checks


mthaddon (approved):
LGTM

PR #89, opened 2022-08-26
Increase SQLAlchemy connection pool size and reduce timeouts

arturo-seijas:
- Reduce connection timeout to 5s
- Increase pool size to 20


mthaddon (approved):
Thx

PR #91, opened 2022-08-26
Fix indico health check

arturo-seijas:
Fix indico health check bug introduced in #88


mthaddon (approved):
Ack, does seem like we need this.

PR #97, opened 2022-08-29
Set http proxy env vars in indico containers

arturo-seijas:
Set http proxy env vars in indico containers to prevent timeouts when accessing external resources from the application


mthaddon (approved):
LGTM

PR #100, opened 2022-08-30
Fix release workflow

arturo-seijas:
Fix dependencies between integration tests and image publishing introduced by #83


mthaddon (approved):
LGTM, thx

PR #103, opened 2022-08-30
Improve pipeline performance

arturo-seijas:
Some improvements in the pipeline and docker images:

- Remove image caching, as images are now only build once
- Split by docker layers for easier upload to mitigate the current charmhub issues


mthaddon (approved):
Looks great. Really looking forward to seeing some smaller image artifacts at the end of this :)

PR #105, opened 2022-09-01
Export images in docker format

arturo-seijas:
Docker image format is built faster and the current pipeline import/export and upload steps support it.


mthaddon (approved):
LGTM, thx

PR #109, opened 2022-09-01
Export roombooking configuration and disable it by default

arturo-seijas:
Export roombooking plugin configuration and disable it by default.


merkata (approved):
Very clean PR, changes are easy to understand (and they are tested). Nice!

varshigupta12:
LGTM

files/etc/indico/indico.conf
```diff
@@ -10,7 +10,7 @@ CUSTOMIZATION_DIR = '/srv/indico/custom'
 CUSTOMIZATION_DEBUG = os.environ.get('INDICO_CUSTOMIZATION_DEBUG', False)
 DEFAULT_LOCALE = os.environ.get('INDICO_DEFAULT_LOCALE', 'en_GB')
 DEFAULT_TIMEZONE = os.environ.get('INDICO_DEFAULT_TIMEZONE', 'UTC')
-ENABLE_ROOMBOOKING = True
+ENABLE_ROOMBOOKING = os.environ.get('ENABLE_ROOMBOOKING', False)
```
jdkandersson, line 13: This could be nitpicking, since the default in the configuration is already `false`, do you think having defaults in multiple places could be confusing? Eg., what if it was changed to `true` in the future, we would have to remember to update it in 2 locations

jdkandersson (approved):

files/etc/indico/indico.conf
mthaddon: I think the reason it's defaults to `False` here is that per the Indico docs the default is `False` so it's best for us to match upstream there. I also think it's better to have a default that not, so this makes sense to me.

https://docs.getindico.io/en/latest/config/settings/#ENABLE_ROOMBOOKING

mthaddon (commented):
The code changes look good to me, just a question about whether we need to restart indico to apply these changes if someone updates the configuration?

jdkandersson (approved):

arturo-seijas:
> The code changes look good to me, just a question about whether we need to restart indico to apply these changes if someone updates the configuration?

Yes, services will be replanned, but they container won't be restarted

PR #113, opened 2022-09-02
Default to no reply email if no valid sender

arturo-seijas:
Default to no reply email if no valid sender is specified when emaling.


mthaddon (approved):
Great, thanks

PR #117, opened 2022-09-05
Remove CSP header

### LINKED ISSUE #116: PR 108 will likely break staging and production
We landed https://github.com/canonical/indico-operator/pull/108 to address security settings that were flagged using an external tool, however, I believe the way we've implemented it will break current staging and production, as I believe we're only allowing resources from the site itself to be served, and I've noticed as part of troubleshooting another issue that we're loading assets from fonts.google.com as well as assets.ubuntu.com.

We should review which of these settings we need to revert, and figure out a strategy to test these kinds of settings changes before landing further updates.

arturo-seijas:
Follow up #108 
Removes CSP header

Closes #116

### DIFF
--- a/files/etc/nginx/nginx.conf
+++ b/files/etc/nginx/nginx.conf
@@ -29,8 +29,6 @@ http {
     }
 
     location / {
-      add_header X-Content-Type-Options 'nosniff';
-      add_header X-Frame-Options 'SAMEORIGIN';
       root /var/empty/nginx;
       proxy_pass http://localhost:8081;
       proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
--- a/files/etc/nginx/nginx.conf
+++ b/files/etc/nginx/nginx.conf
@@ -29,9 +29,6 @@ http {
     }
 
     location / {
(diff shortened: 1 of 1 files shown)


files/etc/nginx/nginx.conf
```diff
@@ -31,7 +31,7 @@ http {
     location / {
       # Missing CSP headers reported in https://github.com/indico/indico/issues/5486.
       # It can be removed once fixed upstream in Indico.
-      add_header Content-Security-Policy 'self';
+      add_header Content-Security-Policy 'self,*.googleapis.com,*.gstatic.com,*.ubuntu.com,*.canonical.com';
```
mthaddon, line 34: I'm not sure we should be adding these into the charm, as they won't be used by everyone, and if someone uses some other customization configuration these won't be relevant. I think we should just remove the Content Security Policy entry for now until we can determine a good approach here.

arturo-seijas: Indeed

mthaddon (approved):
LGTM, we should add a task to the backlog to revisit this, including the things we've learned about it.

PR #118, opened 2022-09-05
Fix activation for external plugins

arturo-seijas:
Fix issue causing plugins not to be enabled on certain scenarios when active plugins list was not being passed to pebble.


mthaddon (approved):
LGTM, thx

PR #119, opened 2022-09-05
Increase celery readiness check period

arturo-seijas:
Increase celery readiness check period and timeout to mitigate the effect of existing blocking tasks when checking the status.


mthaddon (approved):
LGTM

PR #122, opened 2022-09-06
Enable indico user

arturo-seijas:
The crontab of the user "indico" won't run if disabled. This PR enables it.


mthaddon (approved):
LGTM, thx

PR #124, opened 2022-09-06
Disable caching for non static resources

### LINKED ISSUE #123: Authentication information is cached and displayed for not logged in users
### Bug Description

Authentication information is cached and displayed for not logged in users

### To Reproduce

I'm not sure how easily reproducible this is, but I just opened indico in a private window and was greeted with what looked like I was logged in with Mauro's account, with admin permissions. On staging it seems I was logged in with Varshi's account.

Clicking on edit event takes me to U1, so it doesn't seem like I actually had admin, but it is somewhat concerning that apparently some personal information is also cached.

image

### Environment

events.canonical.com, events.staging.canonical.com

### Relevant log output

```shell
See above
```

### Additional context

_No response_

arturo-seijas:
Closes #123

### DIFF
--- a/files/etc/nginx/nginx.conf
+++ b/files/etc/nginx/nginx.conf
@@ -29,6 +29,7 @@ http {
     }
 
     location / {
+      add_header Cache-Control 'no-store, no-cache';
       root /var/empty/nginx;
       proxy_pass http://localhost:8081;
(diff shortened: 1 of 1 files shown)


mthaddon (approved):
LGTM

PR #126, opened 2022-09-07
Automatically update external plugins

### LINKED ISSUE #120: Prod and Staging websites are not updating customization assets upon new commits
### Bug Description

I was told that Prod and Staging would be checking for new updates to the canonical-indico-customization-files repo, and would pull in new changes every 2-5min. That does not appear to be the case, and I have new commits in that repo that are not reflected on either Staging or Prod.

### To Reproduce

1) Push a new commit to https://github.com/canonical/canonical-indico-customization-files
2) Confirm after many minutes that the changes aren't being pulled in

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

arturo-seijas:
Mechanism to update external plugins. As part of this change, the `refresh-external-plugins`action is renamed to `refresh-external-resources`. This change does:
- Periodically update external plugins specified in configuration
- Enable triggering the update though an action

Closes #120


src/charm.py
```diff
@@ -514,17 +515,24 @@ def _download_customization_changes(self, container):
...
+                    ["git", "pull"],
+                    working_dir=INDICO_CUSTOMIZATION_DIR,
+                    user="indico",
+                    environment=self._get_http_proxy_configuration(),
+                )
+                process.wait_output()
+            if self.config["external_plugins"]:
+                logging.debug("Upgrading external plugins %s", self.config["external_plugins"])
+                plugins = self.config["external_plugins"].split(",")
+                self._install_plugins(container, plugins)
```
mthaddon, line 535: See https://juju.is/docs/sdk/actions - we should be using `event.log` and `event.set_results` to log output to the user and return something based on success or failure.

mthaddon (changes requested):
One comment inline. I like the simplification of switching to updates on the `update-status` hook execution as well.

src/charm.py
arturo-seijas: The current code already returns the proper results. The exec method will throw an exception in not executed successfully:
```
juju run-action indico/0 refresh-external-resources --wait
unit-indico-0:
  UnitId: indico/0
  id: "6"
  message: exit status 1
  results:
    ReturnCode: 1
  status: failed
  timing:
    completed: 2022-09-07 12:58:39 +0000 UTC
    enqueued: 2022-09-07 12:58:37 +0000 UTC
    started: 2022-09-07 12:58:39 +0000 UTC
```
Note that the function is run from both the update_status event and the action, so changing it will be non trivial.

mthaddon: If I just deploy the charm and then run this action I get the following, which doesn't really tell me very much:
```
$ juju run-action --wait indico/0 refresh-customization-changes
unit-indico-0:
  UnitId: indico/0
  id: "2"
  results: {}
  status: completed
  timing:
    completed: 2022-09-07 13:35:18 +0000 UTC
    enqueued: 2022-09-07 13:35:17 +0000 UTC
    started: 2022-09-07 13:35:18 +0000 UTC
```
It hasn't told me (for instance) that there's no `customization_sources_url` option set, so there's nothing to do. I understand reworking it will require a little more work, but I think it's better to have clear output from the action.

arturo-seijas: What is that you're actually suggesting? Add in the results section what was executed? And in this scenario, what would you expect to be returned? We currently have 4 successful scenarios:
- Nothing executed
- Customization files pulled
- Plugins upgraded
- Customization files pulled and plugins upgraded

mthaddon: I think the action should include some kind of output that informs the user of what's happened (or not), and if it fails what the failure was (failed to update plugins, failed to update customization files or both).

arturo-seijas: Pushed some changes to make it look like this
```
unit-indico-0:
  UnitId: indico/0
  id: "14"
  results:
    customization-changes: "True"
    plugin-updates: '[''git+https://github.com/canonical/canonical-indico-themes.git'']'
  status: completed
  timing:
    completed: 2022-09-07 15:24:59 +0000 UTC
    enqueued: 2022-09-07 15:24:52 +0000 UTC
    started: 2022-09-07 15:24:55 +0000 UTC
```

mthaddon (approved):
LGTM, thx

PR #127, opened 2022-09-07
Disable system notices

arturo-seijas:
Disables the setting [SYSTEM_NOTICES_URL](https://docs.getindico.io/en/stable/config/settings/?highlight=sqlalchemy#SYSTEM_NOTICES_URL)


mthaddon (approved):
LGTM

PR #129, opened 2022-09-08
Disable update-status hook

arturo-seijas:
Disable the update-status hook to mitigate [pebble issues](https://bugs.launchpad.net/juju/+bug/1989004) when running on a performance degraded cluster.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@
 branch = true
 
 [tool.coverage.report]
-fail_under = 92
+fail_under = 86
 show_missing = true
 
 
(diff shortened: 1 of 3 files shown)


mthaddon (approved):
LGTM, thx

pyproject.toml
```diff
@@ -6,7 +6,7 @@
 branch = true
 
 [tool.coverage.report]
-fail_under = 92
+fail_under = 86
```
gtrkiller, line 9: Is there a particular reason why we need to decrease coverage? It doesn't sound like a good idea

arturo-seijas: The code is still being executed by the action and harness doesn't provide a way to trigger it. The test still exists but is skipped, since exercising the code requires the update-status listener to be "enabled"

gtrkiller (approved):

weiiwang01 (approved):

PR #131, opened 2022-09-09
Delete PR comments from github-actions bot when rerunning tests

arturo-seijas:
Deletes the PR comments from github-actions bot when rerunning tests. Comments from other users will be kept.

### DIFF
--- a/.github/files/comment_pr.js
+++ b/.github/files/comment_pr.js
@@ -18,7 +18,7 @@ module.exports = async ({github, context}) => {
         repo: context.repo.repo,
         issue_number: issue_number,
     });
-    await createComment("RES" + JSON.stringify(comments))
+    await createComment("RES" + comments.data)
     // for (const comment of comments) {
     //     if (comment.user.login == 'github-actions') {
     //         await github.rest.issues.deleteComment({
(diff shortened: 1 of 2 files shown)


.github/files/comment_pr.js
```diff
@@ -3,22 +3,41 @@ module.exports = async ({github, context}) => {
...
+
+    const deleteGithubActionsComments = async () => {
+        const comments = await github.rest.issues.listComments({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            issue_number: issue_number,
+        });
+        const githubActionsComments = comments.data.filter(comment => comment.user.login == 'github-actions[bot]')
+        for (const comment of githubActionsComments) {
+            await github.rest.issues.deleteComment({
```
jdkandersson, line 24: Is there a way to send all the requests and await all of them? It could speed up the deletion since it would be done in parallel

jdkandersson (approved):

merkata (approved):
Cool addition, approving.

.github/files/comment_pr.js
arturo-seijas: I didn't consider it necessary, since only one message is posted by the bot.

mthaddon (approved):
LGTM, thx

PR #133, opened 2022-09-13
Cache improvements

arturo-seijas:
Cache improvements for static files server from the application server:

- Custom static files will be cached for 1 minute
- No cache for the rest of the assets
- Private cache where appropriate

### DIFF
--- a/files/etc/nginx/nginx.conf
+++ b/files/etc/nginx/nginx.conf
@@ -38,6 +38,7 @@ http {
       proxy_pass http://localhost:8081;
 
       location ~ ^/(assets/i18n|static/custom)/(.*)$ {
+        expires 1m;
         proxy_hide_header Cache-Control;
         add_header Cache-Control 'public';
         proxy_pass http://localhost:8081;
--- a/files/etc/nginx/nginx.conf
+++ b/files/etc/nginx/nginx.conf
@@ -32,6 +32,8 @@ http {
       root /var/empty/nginx;
       proxy_hide_header Cache-Control;
       add_header Cache-Control 'no-cache,private';
+      add_header X-Content-Type-Options 'nosniff';
+      add_header X-Frame-Options 'SAMEORIGIN';
       proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
       proxy_set_header X-Forwarded-Host $http_host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


gtrkiller (dismissed):
LGTM

mthaddon (commented):
Could you include more detail in the description about what this will mean and why we're doing it?

I'd be interested to know why we're treating `/assets/i18n/` different to /assets/ in general. Also, are we sure we want to cache `/static/custom/` as that will mean customisation changes won't get applied immediately?

files/etc/nginx/nginx.conf
```diff
@@ -29,12 +29,25 @@ http {
     }
 
     location / {
-      add_header Cache-Control 'no-store, no-cache';
       root /var/empty/nginx;
-      proxy_pass http://localhost:8081;
+      proxy_hide_header Cache-Control;
+      add_header Cache-Control 'no-cache,private';
```
weiiwang01, line 34: > There could be several add_header directives. These directives are inherited from the previous level if and only if there are no add_header directives defined on the current level.

`add_header` will override any `add_header` in the previous level, so this directive will make the following top level `add_header` directives ineffective.
```
  add_header X-Content-Type-Options 'nosniff';
  add_header X-Frame-Options 'SAMEORIGIN';
```

This [blog post](https://www.peterbe.com/plog/be-very-careful-with-your-add_header-in-nginx) and [nginx document](https://nginx.org/en/docs/http/ngx_http_headers_module.html) has more information about this.

weiiwang01: This also applies to other inner level `add_header` directives.

files/etc/nginx/nginx.conf
```diff
@@ -29,12 +29,26 @@ http {
...
       proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
       proxy_set_header X-Forwarded-Host $http_host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_pass http://localhost:8081;
+
+      location ~ ^/static/custom/(.*)$ {
+        expires 1m;
```
mthaddon, line 41: I think this is expiring this nginx instance's cache in 1 minute, isn't it? Whereas what's actually happening in staging and production is that we're setting cache control headers on content that's then being cached by an external service for some period of time. Let me know if I'm misunderstanding things.

files/etc/nginx/nginx.conf
arturo-seijas: Indeed. Thanks

weiiwang01 (approved):
LGTM, thanks!

gtrkiller (approved):

mthaddon (approved):
LGTM, thx. Will obviously need some fairly careful testing in staging.

PR #134, opened 2022-09-16
Remove git proxy configuration

arturo-seijas:
Remove the git proxy configuration. It is picking up the environmental variables: https://github.com/pypa/pip/issues/11405


gtrkiller (approved):
LGTM, thx

merkata (approved):
Thanks for the link, all clear. Approving as well.

PR #136, opened 2022-09-21
Switch redis relations to start with cache or broker so we can deploy aliases

mthaddon:
With this change you can deploy the cache or broker with different aliases such as `redis-cache2` or `redis-broker-new`. If you're in a situation where you can't redeploy a juju application with the same name as before, this gives you the flexibility you might need.


arturo-seijas (dismissed):
LGTM

arturo-seijas (approved):
LGTM

merkata (approved):
Great addition, approving.

PR #139, opened 2022-09-21
Access redis relation data directly

arturo-seijas:
Access redis relation data directly from the relation instead of using the provided stored stated to prevent errors when the stored state has not been updated due to the delay when triggering the `redis_relation_changed` event.


mthaddon (approved):
Looks great, thanks.

merkata (approved):
Very nice!

PR #140, opened 2022-09-27
Display indico version in juju status

### LINKED ISSUE #138: We should display the deployed version of Indico in juju application status
### Enhancement Proposal

If you deploy indico currently you get the following as part of `juju status` output:
```
App     Version  Status   Scale  Charm   Channel  Rev  Address         Exposed  Message
indico           waiting    0/1  indico  stable    73  10.152.183.192  no       installing agent
```
We should also include the version of Indico that's running in the relevant field above.

arturo-seijas:
Closes #138

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,6 +6,7 @@
 """Charm for Indico on kubernetes."""
 import logging
 import os
+from re import findall
 from typing import Dict, Tuple
 from urllib.parse import urlparse
 
@@ -180,6 +181,7 @@ class IndicoOperatorCharm(CharmBase):
         self.unit.status = MaintenanceStatus("Starting {} container".format(container.name))
         container.pebble.replan_services()
         if self._are_pebble_instances_ready():
+            self.unit.set_workload_version(self._get_indico_version())
             self.unit.status = ActiveStatus()
         else:
             self.unit.status = WaitingStatus("Waiting for pebble")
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -471,6 +473,13 @@ def _install_plugins(self, container, plugins):
             )
             process.wait_output()
 
+    def _get_indico_version(self):
+        container = self.unit.get_container("indico")
+        process = container.exec(["indico", "--version"])
+        version_string, _ = process.wait_output()
+        version = findall(r"[0-9.]+", version_string)
```
weiiwang01, line 480: A little picky comment, I think raw string is unnecessary here.

weiiwang01 (dismissed):

src/charm.py
arturo-seijas: You're right

src/charm.py
```diff
@@ -471,6 +473,13 @@ def _install_plugins(self, container, plugins):
             )
             process.wait_output()
 
+    def _get_indico_version(self):
+        container = self.unit.get_container("indico")
+        process = container.exec(["indico", "--version"])
```
mthaddon, line 478: Can we use a full path here?

arturo-seijas: We can

mthaddon (approved):
LGTM, thx

jdkandersson (approved):

PR #144, opened 2022-09-28
Configure bandit in pyproject.toml instead of using .bandit file

### LINKED ISSUE #141: Configure bandit in pyproject.toml instead of using separate .bandit config file.
### Enhancement Proposal

Configuration of [bandit](https://github.com/PyCQA/bandit/) could be done in `pyproject.toml` as per https://bandit.readthedocs.io/en/latest/config.html. This would be cleaner as we already use a `pyproject.toml` in our charm repos and we don't need to support an additional config file (`.bandit` at top of repo).

arturo-seijas:
Closes #141

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,4 +1,0 @@
-assert_used:
-  skips: ["*/*test.py", "*/test_*.py"]
-exclude_dirs:
-  - '/venv/'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,11 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+[tool.bandit]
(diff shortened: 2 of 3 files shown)


merkata (commented):
Nice, did the changes work locally? I remember I had issues with configuring bandit, but that could've been just me.

arturo-seijas:
> Nice, did the changes work locally? I remember I had issues with configuring bandit, but that could've been just me.

Yes, I just had to install toml to make it work

weiiwang01 (approved):
LGTM, thanks!

mthaddon (approved):
LGTM, thx

merkata (commented):
Awesome, approving!

jdkandersson:
Nice!

PR #146, opened 2022-10-04
ISD-192 test git hubs container registry performance for existing workflows

merkata:
This PR substitutes using artifacts of the built container images (indico and nginx-indico) that are used between steps with using ghcr.io (GitHub Container Registry). The images are unique per workflow run (with using the run ID as an image tag), a re-run of the same workflow does not change the run ID. The performance is better in terms of shaving off a couple of minutes, as artifacts are not created, then not having to be downloaded in other steps. The setup of a local microk8s registry where the artifacts are loaded is also not needed. Data transfers between Github actions and GHCR are free of charge, also GHCR is free for public repos. Authentication is via GITHUB_TOKEN, which is fed into the containerd configuration file (a docker login action step does not help). The containerd service is restarted in order for the configuration to take effect (as per docmentation). Note that restarting the whole microk8s breaks the test, as the microk8s cluster is not accessible (I haven't looked deeper into that issue and I think restarting the part we want restarted is better than restarting all components).


arturo-seijas (changes requested):
Note that the artifacts where being downloaded [here](https://github.com/canonical/indico-operator/blob/main/.github/actions/publish_oci_resource/action.yaml#L19), so will need to change that action to publish the images that now live in the repository.
Note also that the action `./.github/actions/build_oci_image`is not longer needed and can be removed.

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,64 @@ on:
   pull_request:
   workflow_call:
 
+env:
+  REGISTRY: ghcr.io
+  OWNER: canonical
```
arturo-seijas, line 9: Will this work from a forked repository outside the canonical organization?

arturo-seijas (changes requested):

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,64 @@ on:
...
-          resource-name: indico-image
+          context: .
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_IMAGE_NAME }}:${{ github.run_id }}
+          file: indico.Dockerfile
+
   build-indico-nginx-image:
     name: Build nginx image
     runs-on: ubuntu-20.04
     outputs:
```
arturo-seijas, line 43: Are we still using this output?

arturo-seijas (changes requested):

.github/workflows/integration_test.yaml
merkata: No, it won't. If we want this to work with a forked repo, we'd have to think of how to support it. My opinion is that this stays as is, since forking the repo would usually mean someone suggesting a change here and with a PR against the repo we will see the result.

.github/workflows/integration_test.yaml
merkata: Doesn't seem like it, deleted. Thanks!

.github/workflows/integration_test.yaml
merkata: Still, it does make sense to have it configurable, so I've changed it to use ${{ github.repository_owner }} - really appreciate the feedback!

merkata:
> Note that the artifacts where being downloaded [here](https://github.com/canonical/indico-operator/blob/main/.github/actions/publish_oci_resource/action.yaml#L19), so will need to change that action to publish the images that now live in the repository. Note also that the action `./.github/actions/build_oci_image`is not longer needed and can be removed.

Thanks, I see the value in having one approach for promoting images to charmhub and for using them in integration tests, however I suggest we keep them separate for now and leave it as is. The PR and changes in integration tests show an improvement that is easy to follow, there are small changes. Refactoring everything would require more time and would mix business logic (testing, promoting...) and would result in a bigger PR. I suggest coming up with a follow up story to address your feedback, as it is really good and it needs to be discussed.

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,60 @@ on:
...
-    outputs:
-      image-id: ${{ steps.build.outputs.imageid }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/build_oci_image
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
```
jdkandersson, line 23: Would it be better to use a tag rather than a specific commit for the version of the action? E.g., `docker/login-action@v2.0.0`

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,60 @@ on:
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push indico image
         id: build
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
```
jdkandersson, line 31: Similar to previous comment, would it be better to use a version tag than a specific commit for the version?

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,60 @@ on:
...
-    outputs:
-      image-id: ${{ steps.build.outputs.imageid }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/build_oci_image
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
```
jdkandersson, line 46: Same as above re targeting commit versus version tag

.github/workflows/integration_test.yaml
```diff
@@ -4,31 +4,60 @@ on:
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push nginx indico image
         id: build
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
```
jdkandersson, line 54: Same as above re targeting commit versus version tag

.github/workflows/integration_test.yaml
```diff
@@ -40,27 +69,16 @@ jobs:
...
-          imageId=$(docker load --input indico-image/indico-image.tar | sed -e "s/^Loaded image ID: //")
-          docker tag $imageId localhost:32000/indico:latest
-          docker push localhost:32000/indico:latest
-      - name: Push nginx image to microk8s registry
+      - name: Configure GHCR in microk8s
         run: |
-          imageId=$(docker load --input indico-nginx-image/indico-nginx-image.tar | sed -e "s/^Loaded image ID: //")
-          docker tag $imageId localhost:32000/indico-nginx:latest
-          docker push localhost:32000/indico-nginx:latest
+          sudo su -c 'echo "
```
jdkandersson, line 74: Do you think this would benefit from some comments explaining what the lines do?

jdkandersson (changes requested):
Thanks for putting this together! It looks like some tests are failing though...

arturo-seijas:
> > Note that the artifacts where being downloaded [here](https://github.com/canonical/indico-operator/blob/main/.github/actions/publish_oci_resource/action.yaml#L19), so will need to change that action to publish the images that now live in the repository. Note also that the action `./.github/actions/build_oci_image`is not longer needed and can be removed.
> 
> Thanks, I see the value in having one approach for promoting images to charmhub and for using them in integration tests, however I suggest we keep them separate for now and leave it as is. The PR and changes in integration tests show an improvement that is easy to follow, there are small changes. Refactoring everything would require more time and would mix business logic (testing, promoting...) and would result in a bigger PR. I suggest coming up with a follow up story to address your feedback, as it is really good and it needs to be discussed.

Unfortunately, we are not uploading the artifacts anymore so the "Release to edge" workflow will fail. If you take a look here you'll se the uploaded images are the ones produced during integration testing: https://github.com/canonical/indico-operator/blob/main/.github/workflows/on_push.yaml#L53

.github/workflows/integration_test.yaml
arturo-seijas: Let keep it as it then. Thanks!

.github/workflows/integration_test.yaml
merkata: Definetely, thanks - added.

.github/workflows/integration_test.yaml
merkata: > > > Note that the artifacts where being downloaded [here](https://github.com/canonical/indico-operator/blob/main/.github/actions/publish_oci_resource/action.yaml#L19), so will need to change that action to publish the images that now live in the repository. Note also that the action `./.github/actions/build_oci_image`is not longer needed and can be removed.
> > 
> > 
> > Thanks, I see the value in having one approach for promoting images to charmhub and for using them in integration tests, however I suggest we keep them separate for now and leave it as is. The PR and changes in integration tests show an improvement that is easy to follow, there are small changes. Refactoring everything would require more time and would mix business logic (testing, promoting...) and would result in a bigger PR. I suggest coming up with a follow up story to address your feedback, as it is really good and it needs to be discussed.
> 
> Unfortunately, we are not uploading the artifacts anymore so the "Release to edge" workflow will fail. If you take a look here you'll se the uploaded images are the ones produced during integration testing: https://github.com/canonical/indico-operator/blob/main/.github/workflows/on_push.yaml#L53

I get it - so I've removed the build (artifact) step in the integration test and now the publish will fail (as there is nothing built), right? So with the `on_push` workflow, we always run tests and I can pass the names of the created images during the run to the `publish_` action and that should do the trick.

arturo-seijas: Exactly, you just need to remove the artifact download and specify the image in [here](https://github.com/canonical/indico-operator/blob/main/.github/actions/publish_oci_resource/action.yaml#L30) instead

arturo-seijas:
Seems like we're almost there. It looks  great.

Do younthink you can remove https://github.com/canonical/indico-operator/tree/main/.github/actions/build_oci_image?

arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test.yaml
merkata: So I checked where I have that from - the docker action repo uses version tags in their examples, whereas github docs make use of SHA and have a comment on all examples:

```yaml
# GitHub recommends pinning actions to a commit SHA.
# To get a newer version, you will need to update the SHA.
# You can also reference a tag or branch, but the action may change without warning.
```

Still I changed the reference to the way docker have it in the examples and it is also aligned with how we reference other actions. If we move to SHA, we maybe should discuss and then transition all actions. Thanks for commenting on this!

.github/workflows/on_push.yaml
```diff
@@ -14,6 +14,12 @@ on:
       - main
       - track/**
 
+env:
+  REGISTRY: ghcr.io
+  OWNER: ${{ github.repository_owner }}
+  INDICO_IMAGE_NAME: indico
+  NGINX_INDICO_IMAGE_NAME: indico-nginx
```
mthaddon, line 21: I wonder if this should be `INDICO_NGINX_IMAGE_NAME` to match the fact that it's called `indico-nginx` rather than `nginx-indico`? If we do change that we'd obviously need to make changes below to match.

arturo-seijas (approved):

mthaddon (approved):
LGTM, thx

.github/workflows/integration_test.yaml
jdkandersson: All good, whilst tags can change they generally shouldn't and are usually associated with commits themselves, it is just more readable and obvious what version is being used

jdkandersson (approved):

PR #147, opened 2022-10-05
Enable local identity when no identity providers configured

arturo-seijas:
Fixes an issue causing the users not to be able to register when no identity providers are configured. The feature is now enabled based on the existence of configuration for SSO


mthaddon (approved):
LGTM, thanks

varshigupta12 (approved):
LGTM as well

PR #150, opened 2022-10-07
Refactor workflow logic

arturo-seijas:
This PR refactors the test and comment workflows to better align with the next steps: extracting the comment functionality to and action and create reusable workflows.
As part of this PR the promote charm workflow is also improved, giving it a more user friendly name and drowdowns instead of text boxes to select the origin and destination channels.


merkata (dismissed):
Awesome changes, really clear and the refactoring improves readability (and re-usability)! Approving.

.github/files/extract_test_result.py
```diff
@@ -26,17 +26,26 @@ def no_color(s):
...
-        "reports": reports
-    }
-    os.makedirs("report", exist_ok=True)
+
+    sha = os.environ["GITHUB_EVENT_PULL_REQUEST_HEAD_SHA"]
+    reports = []
+    if not lint_success:
+        reports.append(
+            f"Lint checks failed for {sha}\n"
+            f"```{no_color(lint_output).strip()}\n```"
```
weiiwang01, line 35: A newline is needed immediately after the first triple backticks.
This applies to some following reports too.

```
            f"```\n{no_color(lint_output).strip()}\n```"
```

weiiwang01 (changes requested):

tox.ini
```diff
@@ -74,6 +74,15 @@ deps =
 commands =
     coverage report
 
+[testenv:static]
```
jdkandersson, line 77: I think I saw this in another PR and might not have had a chance to comment there, what is the difference between `static` and `lint`? `bandit` seems to be another linting tool, is it worth just including it in the lint output rather than as a separate command?

jdkandersson (changes requested):

.github/files/extract_test_result.py
arturo-seijas: Done

tox.ini
arturo-seijas: As I see it, linting could be considered a part of the static analysis. We could merge both targets, but I don't see a strong driver for that.

weiiwang01 (approved):

arturo-seijas:
Closed as #146 prevents actions from forks to succeed

PR #152, opened 2022-10-10
ISD-207 reusable GitHub ci stages

arturo-seijas:
Refactors the integration test and publish workflows to provide a general solution. With the current changes every existing docker image in the root directory will be built and pushed to the github registry. The pushed images will we passed as arguments to the integration target as `--[image-name]-image [pushed-image]`


.github/workflows/integration_test.yaml
```diff
@@ -81,7 +68,13 @@ jobs:
           " >> /var/snap/microk8s/current/args/containerd-template.toml'
           sudo su -c 'systemctl restart snap.microk8s.daemon-containerd.service && microk8s status --wait-ready'
       - name: Run integration tests
-        run: tox -e integration -- --model testing --indico-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_IMAGE_NAME }}:${{ github.run_id }} --indico-nginx-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_NGINX_IMAGE_NAME }}:${{ github.run_id }}
+        run: |
+          image_names=$(echo '${{ needs.get-images.outputs.images }}' | jq -c '.[]' | sed 's/\"//g')
```
gregory-schiano, line 72: You can use "jq -cr" and the output will be without double-quote, so you can remove the sed
```suggestion
          image_names=$(echo '${{ needs.get-images.outputs.images }}' | jq -cr '.[]')
```

.github/workflows/integration_test.yaml
```diff
@@ -7,61 +7,48 @@ on:
...
-          context: .
-          push: true
-          tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_IMAGE_NAME }}:${{ github.run_id }}
-          file: indico.Dockerfile
-
-  build-indico-nginx-image:
-    name: Build nginx image
+      - name: Get dockerfiles
+        id: set-images
+        run: echo "::set-output name=images::$(ls *.Dockerfile | sed s/\.Dockerfile// |  jq -Rsc '. / "\n" - [""]')"
```
mthaddon, line 22: This will error if there are no files matching in the root directory of the charm. This is the case for nginx-ingress-integrator (which has no workload defined), as well as discourse-k8s (which defines its Dockerfile in the `image` folder). Is there a way of dealing with those cases?

arturo-seijas: Fixed. Thank you. I hadn't considered that scenario

weiiwang01 (dismissed):
LGTM, thanks!

.github/workflows/integration_test.yaml
```diff
@@ -7,61 +7,48 @@ on:
...
+  build-images:
```
gregory-schiano, line 23: As mentioned in the comment of Tom, would be interesting to skip this job if there are no dockerfile

arturo-seijas: The current approach will output an empty JSON array that is needed for the input matrix in order not to fail. This looks simpler to me.

mthaddon: what about the case where the files are in sub-folders? This will simply ignore these currently, do we just make it a requirement that the Dockerfile is in the main directory, and follows this naming convention?

arturo-seijas: I'd go for a "convention over configuration" approach.We can also a subfolder for the dockerfiles right now if you think that'd be more appropriate.

mthaddon: I'm not sure we do want to standardise on having dockerfiles in the top level directory of the project named `${container-name}.Dockerfile` as an `ls` for instance wouldn't group all the dockerfiles together (I realise this is a bit pedantic, but that is the first thing that sprung to mind when looking at this naming convention). If it is possible to look for subdirectories and do the right thing (I assume build with a context of the directory we find the Dockerfile in?) I think it might be worth the extra effort at this stage.

mthaddon: We already have two charms that work this way (dockerfiles not in the root directory) - discourse-k8s and gunicorn-k8s. So I think it's worth having a solution that will work for them. I agree simplicity is good, I'm just not sure `${container-name}.Dockerfile` in the root directory is the right approach.

arturo-seijas: In the case of gunicorn the build context is root; for discourse, it is the directory in which the image is placed. We can scan images regardless of their location (although it will be more complex) but we'll need to decide which will be the build context for those cases and adapt the charms accordingly.

gregory-schiano: Sorry to insist but skipped jobs are shown but as "skipped"
Like:
image

And unfortunately it seems the job will fail if the get-images return an empty array see:
https://github.com/orgs/community/discussions/27096

So I strongly suggest to add the `if` statement

.github/workflows/integration_test.yaml
```diff
@@ -7,61 +7,48 @@ on:
...
-          context: .
-          push: true
-          tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_IMAGE_NAME }}:${{ github.run_id }}
-          file: indico.Dockerfile
-
-  build-indico-nginx-image:
-    name: Build nginx image
+      - name: Get dockerfiles
+        id: set-images
+        run: echo "::set-output name=images::$(ls *.Dockerfile 2> /dev/null | sed s/\.Dockerfile// |  jq -Rsc '. / "\n" - [""]')"
```
jdkandersson, line 22: I'm not sure if the following is possible, could we include the name of the image in the output? E.g., indico? This might make it easier to read the workflow since you would know exactly which image is being accessed rather than having to check based on the index in an array or something like that

.github/workflows/integration_test.yaml
```diff
@@ -81,7 +68,13 @@ jobs:
           " >> /var/snap/microk8s/current/args/containerd-template.toml'
           sudo su -c 'systemctl restart snap.microk8s.daemon-containerd.service && microk8s status --wait-ready'
       - name: Run integration tests
-        run: tox -e integration -- --model testing --indico-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_IMAGE_NAME }}:${{ github.run_id }} --indico-nginx-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.INDICO_NGINX_IMAGE_NAME }}:${{ github.run_id }}
+        run: |
+          image_names=$(echo '${{ needs.get-images.outputs.images }}' | jq -cr '.[]')
```
jdkandersson, line 72: Could it be worth formatting the output such that further processing isn't required? It would simplify this a bit

jdkandersson (changes requested):

.github/workflows/integration_test.yaml
arturo-seijas: @gregory-schiano , it does indeed fail. I added that ìf ` and an `if always()` condition on the next job to properly handle empty values returned by the first step

.github/workflows/integration_test.yaml
arturo-seijas: We have visibility in the next step. Does that work for you?
image

.github/workflows/integration_test.yaml
gregory-schiano: Not sure it's possible github actions enforce the output format as json while bash can't iterate on json array directly
But we can directly iterate in the for loop from the output of JQ (prehaps less readable)
```suggestion
          for image_name in $(echo '${{ needs.get-images.outputs.images }}' | jq -cr '.[]'); do
```

.github/workflows/integration_test.yaml
gregory-schiano: Thanks, good catch for the "always()" didn't think about the next job

.github/workflows/integration_test.yaml
arturo-seijas: The best I could do is:
```
args=$(echo '${{ needs.get-images.outputs.images }}' | jq -cr '.[]' | \
            xargs -I{} echo "--${image_name}-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${image_name}:${{ github.run_id }}" | \
            paste -sd' ' -
```
Any suggestions?

gregory-schiano: I think the for loop is more readable than the xargs / paste

arturo-seijas: Agreed

jdkandersson (dismissed):

weiiwang01 (approved):
LGTM, thanks!

merkata (approved):
Significantly reduced and improved readability, awesome! Approving.

.github/workflows/publish_charm.yaml
```diff
@@ -0,0 +1,61 @@
...
+          password: ${{ secrets.github-token }}
+      - name: Publish image
+        env:
+          CHARMCRAFT_AUTH:  ${{ secrets.charmhub-token }}
+        run: |
+          image=${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
+          charmName=$(yq '.name' metadata.yaml)
+          docker pull $image
+          imageId=$(docker images $image --format "{{.ID}}")
+          charmcraft upload-resource $charmName ${{ matrix.image }}-image --image=$imageId --verbosity=trace
```
gregory-schiano, line 47: Is verbosity trace needed ? Could lead to too much output ?

arturo-seijas: The upload is a little flaky sometimes. That's why it's in place. Hoepfully this will go away when migrating charmhub to PS5

PR #153, opened 2022-10-11
ISD-203 indico export uwsgi metrics to prometheus

merkata:
This PR configures the uwsgi component of Indico to send statsd statistics and adds a statsd prometheus exporter to consume them and make them available for scraping by prometheus. Unit and integration tests are modified to reflect that as well.


files/etc/indico/uwsgi.ini
```diff
@@ -31,4 +31,6 @@ harakiri-verbose = true
 reload-on-rss = 2048
 evil-reload-on-rss = 8192
 
-offload-threads = 4
\ No newline at end of file
+offload-threads = 4
+
+stats-push = statsd:127.0.0.1:9125
```
arturo-seijas, line 36: Will this publish the metrics in a statsd server in the same container?

arturo-seijas: Following some kind of architectural IoC pattern, we should ideally let prometheus scrape them (although the other approach is also possible but it’s really aimed for workloads such as cronjobs, that can’t expose an endpoint permanently):

Prometheus will scrape the target exposed by the prometheus-statsd exporter

Prometheus statsd exported should scrape the uwsgi statsd target

arturo-seijas: Nvm, apparently this is not supported by the exporter

src/charm.py
```diff
@@ -285,14 +288,36 @@ def _get_nginx_prometheus_exporter_pebble_config(self, _):
...
+                "nginx-exporter-up": {
                     "override": "replace",
                     "level": "alive",
                     "http": {"url": "http://localhost:9113/metrics"},
                 },
             },
         }
 
+    def _get_statsd_prometheus_exporter_pebble_config(self, _):
+        """Generate pebble config for the statsd-prometheus-exporter container."""
```
jdkandersson, line 300: We have a coding standard for docstrings which includes describing return values, do you think it is applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

tests/integration/test_charm.py
```diff
@@ -37,6 +37,20 @@ async def test_indico_is_up(ops_test: OpsTest, app: Application):
     assert response.status_code == 200
 
 
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_prom_exporters_are_up():
+    """Check that the prometheus exporters are up and running.
```
jdkandersson, line 43: We have a standard for describing tests, do you think it is applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

jdkandersson (changes requested):

src/charm.py
merkata: I wasn't sure, as I copied the `_get_nginx_prometheus_exporter_pebble_config` and modified it to my needs, and it didn't have our standard (and at that point no function has it). I can use the standard for the new functions and the old ones can be refactored in an additional story, I think that would be fine - would do you think?

tests/integration/test_charm.py
merkata: Same as above.

src/charm.py
jdkandersson: Yes, that is also in line with our other standard for non-compliant code 😉 https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#non-compliant-code

merkata: Yes, I remember discussing it - thanks for the pointer, I'm adding this now! :)

jdkandersson (dismissed):
Nice!

jdkandersson (dismissed):

arturo-seijas (approved):
LGTM

gtrkiller (approved):
LGTM

PR #156, opened 2022-10-18
Fix dashboard pannels

arturo-seijas:
Variable `instance` is removed given that the relation already creates the equivalent `juju_unit` variable. Query filters are changed accordingly. Fixes an issue that might appear due to conflicting filters (`instance` and `juju_unit`).
Datasources are also removed since the relation library will inject them.


mthaddon (approved):
LGTM, thx

weiiwang01 (approved):

PR #160, opened 2022-10-19
Fix prometheus metrics scrape

arturo-seijas:
Fix prometheus metrics scrape. `localhost` is not working as hostname for the targets #153 
Remove repeated panel as it is incompatible with the injected filters


merkata (approved):
Thanks so much for spotting and fixing!

mthaddon (approved):
LGTM, thx

PR #162, opened 2022-10-20
Fix failing unit tests with ops 1.5.3

gregory-schiano:
Due to a bad usage of "add_relation" method of Harness, the unit tests started to fail with ops 1.5.3 (which is more strict in validating data relations, see https://github.com/canonical/operator/pull/786)


arturo-seijas (approved):
LGTM

mthaddon (approved):
LGTM, thx

PR #166, opened 2022-10-25
Pin juju dependency

arturo-seijas:
Pin juju dependency to avoid https://github.com/juju/python-libjuju/issues/758


mthaddon (approved):
LGTM

gtrkiller (approved):
LGTM

PR #175, opened 2022-11-09
ISD-263 add uWSGI reload to refresh-external-resources action

amandahla:
On refresh-external-resources action, added two commands:

- touch /srv/indico/indico.wsgi (touch-reload defined in uwsgi.ini)

### DIFF
--- a/indico.Dockerfile
+++ b/indico.Dockerfile
@@ -4,7 +4,7 @@ RUN apt update \
     && apt install -y libpq-dev libxmlsec1-dev pkg-config python3-pip
 
 ENV UWSGI_EMBED_PLUGINS=stats_pusher_statsd
-RUN pip install --prefer-binary indico indico-plugin-piwik indico-plugin-storage-s3 python3-saml uwsgi
+RUN pip install --prefer-binary indico==3.2 indico-plugin-piwik indico-plugin-storage-s3 python3-saml uwsgi
 
 FROM ubuntu:jammy as target
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,7 +7,7 @@
 import logging
 import os
 from re import findall
-from typing import Dict, Tuple
+from typing import Dict, List, Tuple
 from urllib.parse import urlparse
 
 import ops.lib
@@ -25,6 +25,7 @@ DATABASE_NAME = "indico"
 INDICO_CUSTOMIZATION_DIR = "/srv/indico/custom"
 PORT = 8080
 UBUNTU_SAML_URL = "https://login.ubuntu.com/saml/"
+UWSGI_TOUCH_RELOAD = "/srv/indico/indico.wsgi"
 
 pgsql = ops.lib.use("pgsql", 1, "postgresql-charmers@lists.launchpad.net")
 
@@ -507,6 +508,22 @@ class IndicoOperatorCharm(CharmBase):
         version = findall("[0-9.]+", version_string)
         return version[0] if version else ""
 
+    def _exec_cmd_in_custom_dir(self, container, command: List[str]):
+        """Execute command in indico customization directory.
+
+        Args:
+                container: container in a unit where the command will be executed
+                command: Command to execute. The first item is the name (or path)
+                of the executable, the rest of the items are the arguments.
+        """
+        process = container.exec(
+            command,
+            working_dir=INDICO_CUSTOMIZATION_DIR,
+            user="indico",
+            environment=self._get_http_proxy_configuration(),
+        )
+        process.wait_output()
+
     def _download_customization_changes(self, container):
         """Clone the remote repository with the customization changes."""
         current_remote_url = self._get_current_customization_url()
@@ -531,13 +548,9 @@ class IndicoOperatorCharm(CharmBase):
                     "New URL repo for customization %s. Cloning contents",
                     self.config["customization_sources_url"],
                 )
-                process = container.exec(
-                    ["git", "clone", self.config["customization_sources_url"], "."],
-                    working_dir=INDICO_CUSTOMIZATION_DIR,
-                    user="indico",
-                    environment=self._get_http_proxy_configuration(),
+                self._exec_cmd_in_custom_dir(
+                    container, ["git", "clone", self.config["customization_sources_url"], "."]
                 )
-                process.wait_output()
 
     def _refresh_external_resources(self, _) -> Dict:
         """Pull changes from the remote repository and upgrade external plugins."""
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -507,6 +509,16 @@ def _get_indico_version(self):
         version = findall("[0-9.]+", version_string)
         return version[0] if version else ""
 
+    def _execute_command_customdir(self, container, command):
+        """Execute command in indico customization directory"""
```
jdkandersson, line 513: We have a standard for docstrings that includes things like describing arguments and return values, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

src/charm.py
```diff
@@ -23,6 +23,8 @@
 
 DATABASE_NAME = "indico"
 INDICO_CUSTOMIZATION_DIR = "/srv/indico/custom"
+UWSGI_TOUCH_RELOAD = "/srv/indico/indico.wsgi"
```
arturo-seijas, line 26: Can we alphabetically order these constants

src/charm.py
```diff
@@ -507,6 +509,16 @@ def _get_indico_version(self):
         version = findall("[0-9.]+", version_string)
         return version[0] if version else ""
 
+    def _execute_command_customdir(self, container, command):
```
arturo-seijas, line 512: Can we rename this to something like `execute_commend_in_custom_dir` or `exec_cmd_in_custom_dir` if the former is too long for a little more clarity?

src/charm.py
```diff
@@ -550,13 +558,13 @@ def _refresh_external_resources(self, _) -> Dict:
...
+                    container,
                     ["git", "pull"],
-                    working_dir=INDICO_CUSTOMIZATION_DIR,
-                    user="indico",
-                    environment=self._get_http_proxy_configuration(),
                 )
-                process.wait_output()
+                logging.debug("Reloading uWSGI")
+                self._execute_command_customdir(container, ["touch", UWSGI_TOUCH_RELOAD])
+                self._execute_command_customdir(container, ["curl", "-sf", UWSGI_HTTP_SOCKET])
```
arturo-seijas, line 567: What's the purpose behind executing this curl? Will it block here waiting for the response for a while or is it just to check that uwsgi is responding?

src/charm.py
amandahla: I think in this case we could opt for the one-line docstring since there are no return values, optional arguments, or default values for them. It's similar to other methods like "_install_plugins".  What do you think?

src/charm.py
amandahla: Sure!

src/charm.py
amandahla: Done.

src/charm.py
amandahla: Just to check if uWSGI reload was successful. If for some reason UWSGI_HTTP_SOCKET is down, curl will fail.

src/charm.py
amandahla: At first, I thought that wouldn't be necessary but then saw here (https://ops.readthedocs.io/en/latest/_modules/ops/pebble.html#Client.exec) about the command argument. So I'll use the same description.

weiiwang01 (dismissed):

merkata (dismissed):
Nicely done! Approving.

src/charm.py
```diff
@@ -507,6 +508,22 @@ def _get_indico_version(self):
         version = findall("[0-9.]+", version_string)
         return version[0] if version else ""
 
+    def _exec_cmd_in_custom_dir(self, container, command: List[str]):
+        """Execute command in indico customization directory.
+
+        Args:
+                container: container in a unit where the command will be executed
```
jdkandersson, line 515: Should this be indented by 4 spaces?

src/charm.py
```diff
@@ -507,6 +508,22 @@ def _get_indico_version(self):
         version = findall("[0-9.]+", version_string)
         return version[0] if version else ""
 
+    def _exec_cmd_in_custom_dir(self, container, command: List[str]):
```
jdkandersson, line 511: Should a return type be defined here?

src/charm.py
```diff
@@ -507,6 +508,22 @@ def _get_indico_version(self):
...
+        """Execute command in indico customization directory.
+
+        Args:
+                container: container in a unit where the command will be executed
+                command: Command to execute. The first item is the name (or path)
+                of the executable, the rest of the items are the arguments.
```
jdkandersson, line 517: Should this be indented a little further than the above to indicate continuation of the description for the `command` parameter?

indico.Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apt update \
     && apt install -y libpq-dev libxmlsec1-dev pkg-config python3-pip
 
 ENV UWSGI_EMBED_PLUGINS=stats_pusher_statsd
-RUN pip install --prefer-binary indico indico-plugin-piwik indico-plugin-storage-s3 python3-saml uwsgi
+RUN pip install --prefer-binary indico==3.2 indico-plugin-piwik indico-plugin-storage-s3 python3-saml uwsgi
```
arturo-seijas, line 7: Apparently adding `--force` to the db scripts will enable us to deploy 3.2.1 https://github.com/indico/indico/pull/5503/files

amandahla: Good: "the latest version we test with is 12"
Not so good: "but we make no guarantees  that nothing is broken" 
;-)
Do you think that we can take the risk? Since 12 it's the version in use now.

src/charm.py
amandahla: This method is very similar with _install_plugins. There is no return.

indico.Dockerfile
arturo-seijas: Sorry, I forgot to mention that the `--force ` option doesn't seem to be working anyway. Let's go with pinning, good points

arturo-seijas (dismissed):
LGTM

weiiwang01 (approved):

src/charm.py
jdkandersson: In that case, the usual return type is `None`: ` def _exec_cmd_in_custom_dir(self, container, command: List[str]) -> None:` which is the default return value in any Python function that doesn't return anything

arturo-seijas (approved):
LGTM

src/charm.py
amandahla: If it's ok we can create a new story to review all docstrings for this charm.

PR #176, opened 2022-11-10
ISD-189 indico chaos engineering

merkata:
This is a draft and a PoC of two possible solutions for chaos testing - Litmus and Chaos Engine. I've included them locally, once we settle for one or the other, the respective code can be opened in our actions repository. A successful run is available at https://github.com/canonical/indico-operator/actions/runs/3431412702

Both Litmus and Chaos Engine actions are not that well supported in my opinion:
for Litmus, I've opened a PR upstream with a fix that is still not addressed two weeks later
for Chaos Engine, the example references Chaos Engine v1.0.0 with v2.4.1 being the latest available, also the master branch has a fix for CRDs that cannot be applied due to the size of the manifest (on `apply` you get a copy of the manifest in the annotations and there is a limitation, you can use `create` instead which does not copy data to the annotations), but there hasn't been any release for a while now

Litmus is better in terms of presenting the result of the test, for Chaos Engine there is no statement that the experiment has run to completion and was successful, so testing in CI/CD doesn't seem suitable.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -118,7 +118,7 @@ jobs:
           EXPERIMENT_NAME: pod-delete
 
       - name: Run chaos mesh pod kill action
-        uses: chaos-mesh/chaos-mesh-action@v0.5
+        uses: chaos-mesh/chaos-mesh-action@master
         env:
           CHAOS_MESH_VERSION: v2.4.1
           CFG_BASE64: YXBpVmVyc2lvbjogY2hhb3MtbWVzaC5vcmcvdjFhbHBoYTEKa2luZDogUG9kQ2hhb3MKbWV0YWRhdGE6CiAgbmFtZTogcG9kLWtpbGwtZXhhbXBsZQogIG5hbWVzcGFjZTogY2hhb3MtbWVzaApzcGVjOgogIGFjdGlvbjogcG9kLWtpbGwKICBtb2RlOiBvbmUKICBzZWxlY3RvcjoKICAgIG5hbWVzcGFjZXM6CiAgICAgIC0gdGVzdGluZwogICAgbGFiZWxTZWxlY3RvcnM6CiAgICAgICdhcHAua3ViZXJuZXRlcy5pby9uYW1lJzogJ2luZGljbycK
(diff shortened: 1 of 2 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -70,17 +84,41 @@ jobs:
...
\ No newline at end of file
+          model: testing
+
+      - name: Setting up kubeconfig ENV for Github Chaos Action
+        run: echo ::set-env name=KUBE_CONFIG_DATA::$(sudo microk8s config | base64 -w 0)
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+
+      - name: Setup Litmus
+        uses: merkata/github-chaos-actions@master
```
jdkandersson, line 109: Should this be in the canonical organisation?

.github/workflows/test.yaml
```diff
@@ -3,6 +3,11 @@ name: Tests
 on:
   workflow_call:
   pull_request:
+  push:
+  branches:
+    - 'feat/linters'
```
jdkandersson, line 8: Out of curiosity, what is this for please?

jdkandersson (commented):
Is there no well supported chaos engineering tool available?

merkata:
> Is there no well supported chaos engineering tool available?

I used the tools that have most support and are part of the CNCF projects, these are Litmus Chaos and Chaos Engine. I think both are actively developed and I also think the corresponding official actions to them are not (see description of the PR).

.github/workflows/integration_test.yaml
merkata: It should be in the `litmuschaos` organization, as they support it, however the fix https://github.com/litmuschaos/chaos-ci-lib/pull/87 that I opened has not received any review yet (18 days later). Without the fix, we're not able to specify that we want to test against a statefulset, rather the default gets picked up and the test runs against a deployment (and fails, as no deployment is found).

.github/workflows/test.yaml
merkata: That creeped in while working on a different feature about integrating hadolint and shellcheck, I was using a test branch on indico and was switching branches. It is here by mistake, thanks for noticing and pointing it out!

.github/workflows/integration_test.yaml
jdkandersson: Do you think it is worth including that information as a comment so we know when we can revert to the upstream library in the future?

merkata:
I'm closing this, as the PoC is done and was presented, next step would be to integrate chaos capabilities in the workflows repo, so that charm repos can use them.

PR #179, opened 2022-11-14
Force deployment with postgresql 12

arturo-seijas:


merkata (approved):
Based on what you provided as info in a different PR (and the link https://github.com/indico/indico/pull/5503/files explaining the change), I'm fine with that. This will go to the staging env, correct?

PR #180, opened 2022-11-14
Test shellcheck annotations

merkata:
I want to see how shellcheck annotations look like with the additional condition of not failing bellow error severity and having warnings - the CI passes, but will there be annotations in the PR?


merkata:
This PR has served its purpose of trying out different things in order to fix annotations on the shellcheck action that is being used. The result of this research is in a PR in the operator-framework so it can be reused across charm repos. Closing this.

PR #184, opened 2022-11-18
Feat/linters - test picking up shebang shell scripts without extension and annotating in PR

merkata:
Testing that files without a .sh extension and with a present shebang are picked up. This currently relies that such files are executable (perm /111) and don't have an extension (! -f *.*) - that is directly taken from the shellcheck action used previously that has shell script detection. We can loosen that up if needed.


merkata:
Files are picked up and annotations are present, closing this and opening another PR in the operator-workflows.

PR #186, opened 2022-11-25
Refactor dockerfiles

arturo-seijas:


jdkandersson (approved):

jdkandersson:
The feedback from the Dockerfile linting could be worth reading through, there were some good suggestions in there

indico-nginx.Dockerfile
```diff
@@ -1,24 +1,27 @@
...
-RUN apt update \
-    && apt install -y libpq-dev python3-pip \
-    && pip install --prefer-binary indico==3.2 indico-plugins
+ARG nginx_gid=2001
+ARG nginx_uid=2001
+
+RUN apt-get update \
+    && apt-get install -y libpq-dev python3-pip \
+    && addgroup --gid ${nginx_gid} nginx \
+    && adduser --system --gid ${nginx_gid} --uid ${nginx_uid} --home /srv/indico --disabled-login nginx
```
gregory-schiano, line 11: I'm wondering if it's useful to add the user/group in this stage as it's not the final stage of the image

indico.Dockerfile
```diff
@@ -23,16 +26,11 @@ RUN apt update \
 RUN /bin/bash -c "mkdir -p --mode=775 /srv/indico/{etc,tmp,log,cache,archive,custom}" \
     && /usr/local/bin/indico setup create-symlinks /srv/indico
```
gregory-schiano, line 27: Could this be done in the builder and then copied from it ? (didn't check what indico setup is doing)

indico.Dockerfile
```diff
@@ -23,16 +26,11 @@ RUN apt update \
...
     && adduser --system --gid ${indico_gid} --uid ${indico_uid} --home /srv/indico indico
 
-COPY files/start-indico.sh /srv/indico/
-COPY files/etc/indico/ /etc/
+COPY --chown=indico:indico files/start-indico.sh /srv/indico/
+COPY --chown=indico:indico files/etc/indico/ /etc/
 
 RUN chmod +x /srv/indico/start-indico.sh \
-    && chown -R indico:indico /srv/indico \
-    && chmod 755 /srv/indico
```
gregory-schiano, line 37: I suppose it was useless since the `mkdir` creates parent directory and sets to mode `755` the created folder ?

merkata (approved):

indico.Dockerfile
arturo-seijas: Exactly. I've added a `chmod`earlier that will make this one useless also

indico.Dockerfile
arturo-seijas: It does create some symbolic links and copies some files from the Python package to other locations

indico-nginx.Dockerfile
arturo-seijas: I've had some troubled installing the packages as root in the Indico image, so I decided to follow recommendations and install as non toot here as well

PR #187, opened 2022-11-25
Ldap support

arturo-seijas:


jdkandersson (dismissed):

src/charm.py
```diff
@@ -609,7 +630,48 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
+                    "ubuntu_ldap": {
+                        "type": "ldap",
+                        "title": "LDAP",
+                        "ldap": _ldap_config,
+                        "mapping": {
+                            "email": "mail",
+                            "affiliation": "o",
+                            "first_name": "givenName",
+                            "last_name": "sn",
+                            "phone": "mobile",
```
mthaddon, line 660: Are we really planning to sync phone numbers? Is that required by Indico somehow? Seems potentially problematic or unexpected.

src/charm.py
```diff
@@ -609,7 +630,48 @@ def _get_indico_env_config(self, container: Container) -> Dict:
                     "identifier_field": "openid",
                 }
             }
+            if self.config["ldap_host"]:
+                _ldap_config = {
```
mthaddon, line 634: It seems odd that we're pulling some variables here (`ldap_host`) and hard coding others (`cn=Indico Bot,ou=bots,dc=canonical,dc=com`). I realise we're only supporting one provider currently, but surely this should be checking if the host matches what we're expecting and mapping values to that? Maybe just `if self.config["ldap_host"] == "ldap.canonical.com"`?

.github/workflows/integration_test.yaml
```diff
@@ -12,3 +12,4 @@ jobs:
       chaos-experiments: pod-delete
       chaos-app-label: app.kubernetes.io/name=indico
       load-test-enabled: true
+      load-test-run-args: "-e LOAD_TEST_HOST=localhost"
```
mthaddon, line 15: Is this intended, seems unrelated?

config.yaml
```diff
@@ -38,6 +38,14 @@ options:
     type: string
     description: HTTPS proxy URL to be used when installing plugins or downloading customization files, e.g. 'http://squid.example:3128'.
     default: ''
+  ldap_host:
+    type: string
+    description: URL for the LDAP server to use as identity provider. Requires SAML to be configured.
+    default: ''
+  ldap_password:
+    type: string
+    description: LDAP for the LDAP server to use as identity provider.
```
mthaddon, line 47: This description doesn't quite make sense. Can you reword it?

config.yaml
```diff
@@ -38,6 +38,14 @@ options:
     type: string
     description: HTTPS proxy URL to be used when installing plugins or downloading customization files, e.g. 'http://squid.example:3128'.
     default: ''
+  ldap_host:
+    type: string
+    description: URL for the LDAP server to use as identity provider. Requires SAML to be configured.
```
gregory-schiano, line 43: Is it and URI or a host ?
If that's an URI I'd call it `ldap_uri`

src/charm.py
arturo-seijas: Not sure I follow. I see only two variable in that block: ldap_host and ldap_password

config.yaml
gregory-schiano: Yeap he probably meant `Password for the LDAP server`

arturo-seijas: Done

.github/workflows/integration_test.yaml
arturo-seijas: It is, we have some checks failing because of that, though we will probably remove this check on PR submission

indico.Dockerfile
```diff
@@ -20,14 +26,17 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
-
-RUN addgroup --gid ${indico_gid} indico \
-    && adduser --system --gid ${indico_gid} --uid ${indico_uid} --home /srv/indico indico
+RUN apt-get update \
+    && apt-get install -y gettext git libxmlsec1-dev locales postgresql-client python3-pip texlive-xetex \
+    && rm -rf /var/lib/apt/lists/* \
+    && addgroup --gid ${indico_gid} indico \
+    && adduser --system --gid ${indico_gid} --uid ${indico_uid} --home /srv/indico indico \
+    && /bin/bash -c "mkdir -p --mode=775 /srv/indico/{archive,cache,custom,etc,log,tmp}" \
+    && chown indico:indico /srv/indico  /srv/indico/cache /srv/indico/archive /srv/indico/custom \
```
gregory-schiano, line 35: Why not using the {} as in the previous command here ?

src/charm.py
```diff
@@ -478,22 +479,56 @@ def _get_redis_cache_rel(self) -> Optional[Relation]:
...
+        Returns:
+            List containing the installed plugins.
+        """
+        process = container.exec(
+            ["/srv/indico/.local/bin/indico", "setup", "list-plugins"],
+            user="indico",
+        )
+        output, _ = process.wait_output()
+        # Parse output table, discarding header and footer rows and fetching first column value
+        return [item.split("|")[1].strip() for item in output.split("\n")[3:-2]]
```
gregory-schiano, line 530: Did you test with no plugins installed ?

P.S.: Ah I see this was pre-existing code, but could worth it checking this edge case

indico.Dockerfile
```diff
@@ -1,16 +1,22 @@
...
+    && apt-get install -y libldap2-dev libpq-dev libsasl2-dev libssl-dev libxmlsec1-dev pkg-config python3-pip \
+    && rm -rf /var/lib/apt/lists/* \
+    && addgroup --gid ${indico_gid} indico \
+    && adduser --system --gid ${indico_gid} --uid ${indico_uid} --home /srv/indico indico
 
 ENV UWSGI_EMBED_PLUGINS=stats_pusher_statsd
-RUN pip install --prefer-binary indico==3.2 indico-plugin-piwik indico-plugin-storage-s3 python3-saml uwsgi
+
+USER indico
+RUN python3 -m pip install --no-cache-dir --prefer-binary indico==3.2 indico-plugin-piwik indico-plugin-storage-s3 python-ldap python3-saml uwsgi
```
nrobinaubertin, line 15: I'm not completely sure, but I thought that to install to `$HOME/.local`, you needed to add `--user` to the `pip install` command

src/charm.py
gregory-schiano: This charm is not "Canonical" only, so indeed these variables shouldn't be hardcoded as it'll make it unusable for other users.
Usually on application providing LDAP connection provides at least the following configuration params:
- bind DN
- user DN
- user filter
- group DN
- group filter
- And fields to fetch for username, email, id, group_name, ...

src/charm.py
```diff
@@ -609,7 +630,48 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
+                    "user_base": "ou=staff,dc=canonical,dc=com",
+                    "user_filter": "(objectClass=canonicalPerson)",
+                    "gid": "cn",
+                    "group_base": "dc=canonical,dc=com",
+                    "group_filter": "(objectClass=canonicalTeam)",
+                    "member_of_attr": "mozillaCustom1",
+                    "ad_group_style": False,
+                }
+                identity_providers = {
+                    "ubuntu_ldap": {
```
gregory-schiano, line 651: I'd make this naming generic like "ldap_provider"

src/charm.py
```diff
@@ -609,7 +630,48 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
+                            "first_name": "givenName",
+                            "last_name": "sn",
+                            "phone": "mobile",
+                        },
+                        "trusted_email": True,
+                        "synced_fields": {"first_name", "last_name", "affiliation", "phone"},
+                    }
+                }
+                provider_map = {
+                    "ubuntu": {
```
gregory-schiano, line 667: I'd make this name generic like `provider`

src/charm.py
```diff
@@ -647,6 +713,15 @@ def _on_config_changed(self, event: HookEvent) -> None:
             )
             event.defer()
             return
+        if not self._is_ldap_host_valid():
+            self.unit.status = BlockedStatus(
+                f"Invalid ldap_host option provided. Only {CANONICAL_LDAP_HOST} is available."
+            )
+            event.defer()
+            return
```
gregory-schiano, line 721: Why limiting to Canonical LDAP only ?

src/charm.py
```diff
@@ -478,22 +479,56 @@ def _get_redis_cache_rel(self) -> Optional[Relation]:
...
         broker_port = ""
-        if (broker_rel := self._get_redis_broker_rel()) is not None:
-            broker_unit = next(
-                unit for unit in broker_rel.data if unit.name.startswith("redis-broker")
-            )
+        if (broker_rel := self._get_redis_rel(name)) is not None:
+            broker_unit = next(unit for unit in broker_rel.data if unit.name.startswith(name))
             broker_host = broker_rel.data[broker_unit].get("hostname")
             broker_port = broker_rel.data[broker_unit].get("port")
         return f"redis://{broker_host}:{broker_port}"
```
nrobinaubertin, line 497: Since this function is for redis cache and broker, I would avoid using the words broker and cache (and replace them by just "redis")

nrobinaubertin (changes requested):

config.yaml
arturo-seijas: Host, not URI `"uri": f"ldaps://{self.config['ldap_host']}"`

src/charm.py
arturo-seijas: It would be quite complex to pass all these configs, so that's the scope for now, same as for SAML

src/charm.py
arturo-seijas: It won't be reusable anyway

src/charm.py
arturo-seijas: It won't be reusable anyway

indico.Dockerfile
arturo-seijas: Running as the user also does the trick

src/charm.py
arturo-seijas: Indeed. Fixed

src/charm.py
arturo-seijas: It would be quite complex to pass all these configs, so that's the scope for now, same as for SAML

src/charm.py
mthaddon: Yes please, let's remove it.

arturo-seijas: Done

indico.Dockerfile
arturo-seijas: Indeed. Changed

src/charm.py
mthaddon: I agree we should try to avoid having to specify all those values in juju config if possible, and I think the approach of only requiring ldap_host and ldap_password makes sense, but my comment was about the fact that we're checking `if self.config["ldap_host"] but then defining `_ldap_config` to include some hard coded and some parameterised values. If we add support for other LDAP providers in the future (or if someone like me is reading through the code) it might be a bit confusing.

nrobinaubertin (dismissed):
LGTM

src/charm.py
arturo-seijas: Done

src/charm.py
arturo-seijas: Done

src/charm.py
arturo-seijas: Changed

src/charm.py
```diff
@@ -35,6 +35,7 @@
 NGINX_PROMEXP_PORT = "9113"
 STATSD_PROMEXP_PORT = "9102"
 UBUNTU_SAML_URL = "https://login.ubuntu.com/saml/"
+CANONICAL_LDAP_HOST = "ldap.canonical.com"
```
amandahla, line 38: alphabetical order?

src/charm.py
```diff
@@ -626,6 +687,10 @@ def _get_http_proxy_configuration(self) -> Dict[str, str]:
             config["HTTPS_PROXY"] = self.config["https_proxy"]
         return config
 
+    def _is_ldap_host_valid(self) -> bool:
+        """Check if the LDAP hostis currently supported."""
```
amandahla, line 691: Minor thing: host is

src/charm.py
arturo-seijas: Fixed

src/charm.py
arturo-seijas: Fixed. Thanks for your input

jdkandersson (approved):

src/charm.py
gregnebu: As discussed since my comment, we'll keep the implementation as it is (which means limiting to canonical LDAP only) but will work on integrator charms for LDAP / SAML and will refactor this part accordingly

gregory-schiano: As discussed since my comment, we'll keep the implementation as it is (which means limiting to canonical LDAP only) but will work on integrator charms for LDAP / SAML and will refactor this part accordingly

gregory-schiano (approved):
LGTM

PR #195, opened 2022-12-07
docs(ISD-145): Add Explanation

amandahla:
Add an explanation of the charm architecture.


docs/explanation/charm-architecture.md
```diff
@@ -0,0 +1,95 @@
+# Charm Architecture
+
+At its core, [Indico](https://getindico.io/) is a [Flask](https://flask.palletsprojects.com/) application that requires connections to [PostgreSQL](https://www.postgresql.org/), [Redis](https://redis.io/), and [Celery](https://docs.celeryq.dev/en/stable/).
+
+In designing the charm, we've leveraged the [sidecar](https://kubernetes.io/blog/2015/06/the-distributed-system-toolkit-patterns/#example-1-sidecar-containers) pattern to allow us to run multiple containers in each pod with [Pebble](https://juju.is/docs/sdk/pebble) running as the workload container’s entrypoint.
```
jdkandersson, line 5: In the documentation it can be helpful to avoid using `we` or `us` and try to make it neutral. E.g., `The charm design leverages the [sidecar](https://kubernetes.io/blog/2015/06/the-distributed-system-toolkit-patterns/#example-1-sidecar-containers) pattern ...`

docs/explanation/charm-architecture.md
```diff
@@ -0,0 +1,95 @@
...
+
+Pebble is a lightweight, API-driven process supervisor that is responsible for configuring processes to run in a container and controlling those processes throughout the workload lifecycle.
```
jdkandersson, line 7: Would it be worth including a link to more details on pebble?

docs/explanation/charm-architecture.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Celery
+
+The Celery is used to process tasks created by the Indico application such as sending e-mails, survey notifications, event reminders, etc.
+
+The Celery container runs the same workload as the Indico container, as defined in the [indico dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico.Dockerfile).
+
+### NGINX Prometheus Exporter
+
+This container runs the `nginx/nginx-prometheus-exporter` image.
```
jdkandersson, line 54: Would it be worth including a link for more details on the container?

docs/explanation/charm-architecture.md
```diff
@@ -0,0 +1,95 @@
...
+
+The `NGINX Prometheus Exporter` is started with `-nginx.scrape-uri=http://localhost:9080/stub_status` so will scrap metrics from the NGINX container.
+
+This has been configured in the NGINX container to return NGINX's [stub_status](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html). The exporter listens on port `9113` and metrics about web traffic to the pod can be scraped by Prometheus there.
+
+### StatsD exporter
+
+This container runs the `prom/statsd-exporter` image.
```
jdkandersson, line 62: Similar question as the above, possibly a link to the container for more details?

docs/explanation/charm-architecture.md
```diff
@@ -0,0 +1,95 @@
+# Charm Architecture
```
jdkandersson, line 1: Would it be worth also discussing the architecture of the code inside the charm? E.g., key functions? I'm not sure what the answer is, something to consider

docs/explanation/charm-architecture.md
amandahla: I had the same idea and it was really bothering me but in the end, decided to left it after reading the [Google Developer Documentation Style Guide](https://developers.google.com/style/pronouns#personal-pronouns).
In the sense that "we" means "Canonical" and this paragraph tries to justify why the sidecar design pattern was chosen.
But your suggestion looks good too, I'll change it just in case to keep the neutral form.

docs/explanation/charm-architecture.md
amandahla: I was afraid that repeating links could "pollute" the reading. Actually, I'm worried that I already added too many. Usually, I create a link only at the first occurrence of the word. Do you think would be better to put in all of them?
I didn't find a guideline in Diátaxis but Google has even a term for that: "Link fatigue". :open_mouth: 
"To avoid link fatigue, don't provide duplicate links to the same document within a given page. "

docs/explanation/charm-architecture.md
amandahla: Same comment about link fatigue. One link was added at the beginning here:
https://github.com/canonical/indico-operator/blob/c31d156b95feff1b23912ce00251a7e21ad15e24/docs/explanation/charm-architecture.md?plain=1#L14

docs/explanation/charm-architecture.md
amandahla: Same thing as above. Here the first occurrence of the link:
https://github.com/canonical/indico-operator/blob/c31d156b95feff1b23912ce00251a7e21ad15e24/docs/explanation/charm-architecture.md?plain=1#L15

docs/explanation/charm-architecture.md
amandahla: It's very tempting to do that but I think would be out of the scope of the "Explanation" concept. What do you think?
From [Diátaxis](https://diataxis.fr/explanation/):

> It’s not concerned with what the user might be doing, like tutorials and how-to guides. It’s not a close-up view of the machinery, like reference material. It’s documentation that approaches a topic from a higher perspective, and from different angles.

docs/explanation/charm-architecture.md
jdkandersson: Yeh duplicate links are unnecessary. Although making words clickable multiple times can be helpful so the reader doesn't have to find the first reference. Up to you

docs/explanation/charm-architecture.md
jdkandersson: It is a judgement call, the code architecture could still be discussed at a higher level

jdkandersson (dismissed):

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
 
-As mentioned above, this container is the entry point for all web traffic to the pod (on port `8080`), and it serves some static files directly and forwards non-static requests to the indico container (on port `8081`).
+The reason for that is since NGINX provides cache static content, reverse proxy, and load balance among multiple application servers, as well as other features it can be used in front of uWSGI server to significantly reduce server and network load.
 
-## Indico container
+The workload that this container is running is defined in the [NGINX dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico-nginx.Dockerfile).
+
+### Indico
+
+Indico is a Flask application which means a [WSGI](https://wsgi.readthedocs.io/en/latest/what.html) application. A WSGI server is used to run it and uWSGI is the most popular one.
```
arturo-seijas, line 42: I'd rephrase this to something simpler like:
Indico is a Flask run by the [WSGI](https://wsgi.readthedocs.io/en/latest/what.html) server, one of the most popular servers for these applications.

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+
+The uWSGI server is started in HTTP mode (port `8081`) serving Indico Application so NGINX can forward non-static traffic to it.
 
 The workload that this container is running is defined in the [indico dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico.Dockerfile).
 
-On startup, pebble is used to start a uwsgi server in http mode, on port `8081`. Non-static traffic is forwarded here from the nginx container.
+### Celery
+
+The Celery is used to process tasks created by the Indico application such as sending e-mails, survey notifications, event reminders, etc.
```
arturo-seijas, line 50: to process tasks asynchronously

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+
+## Integrations
+
+### Ingress
+
+The Indico charm also supports being integrated with [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#what-is-ingress) by using [NGINX Ingress Integrator](https://charmhub.io/nginx-ingress-integrator/).
+
+In this case, an existing Ingress controller is required. For more information, see [Adding the Ingress Relation to a Charm](https://charmhub.io/nginx-ingress-integrator/docs/adding-ingress-relation).
+
+### PostgreSQL
```
arturo-seijas, line 85: I miss COS here

arturo-seijas (commented):
Some comments inline

arturo-seijas (commented):
I'm wondering what will be the behaviour if the current docs are not defined as source. Will they be deleted? Maybe we should copy all of those that are up to date.

docs/explanation/charm-architecture.md
amandahla: Done!

merkata (dismissed):
Solid foundation, I like it!

docs/explanation/charm-architecture.md
arturo-seijas: Still don't see it

docs/explanation/charm-architecture.md
merkata: It's async, you have to await for it...

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+
+PostgreSQL is an open-source object-relational database used by Indico as a source of all the data needed for its goal: event organization, archival, and collaboration.
 
-## Celery container
+### Redis
```
arturo-seijas, line 89: There are two redis instances that are deployed with different purposes: as cache and as message broker. The later is not mentioned

arturo-seijas (changes requested):
I miss some things, but it looks great overall

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
-## Indico container
+The workload that this container is running is defined in the [NGINX dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico-nginx.Dockerfile).
+
+### Indico
+
+Indico is a Flask application which means a [WSGI](https://wsgi.readthedocs.io/en/latest/what.html) application. A WSGI server is used to run it and uWSGI is the most popular one.
+
+The uWSGI server is started in HTTP mode (port `8081`) serving Indico Application so NGINX can forward non-static traffic to it.
 
 The workload that this container is running is defined in the [indico dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico.Dockerfile).
```
arturo-seijas, line 46: You might want to capitalize `indico`here

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+The reason for that is since NGINX provides cache static content, reverse proxy, and load balance among multiple application servers, as well as other features it can be used in front of uWSGI server to significantly reduce server and network load.
 
-## Indico container
+The workload that this container is running is defined in the [NGINX dockerfile in the charm repository](https://github.com/canonical/indico-operator/blob/main/indico-nginx.Dockerfile).
+
+### Indico
+
+Indico is a Flask application which means a [WSGI](https://wsgi.readthedocs.io/en/latest/what.html) application. A WSGI server is used to run it and uWSGI is the most popular one.
+
+The uWSGI server is started in HTTP mode (port `8081`) serving Indico Application so NGINX can forward non-static traffic to it.
```
arturo-seijas, line 44: You might want to capitalize `indico` here

docs/explanation/charm-architecture.md
arturo-seijas, line 46: You might want to capitalize indico here

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+## Juju Events
 
-## Prometheus-nginx-exporter container
+Accordingly to the [Juju SDK](https://juju.is/docs/sdk/event): "an event is a data structure that encapsulates part of the execution context of a charm".
 
-This container runs the `nginx/nginx-prometheus-exporter` image. 
+For this charm, the following events are observed:
 
-On startup, pebble is configured to run `nginx-prometheus-exporter` with  `-nginx.scrape-uri=http://localhost:9080/stub_status`. This has been configured in the nginx container to return nginx's [stub_status](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html). The exporter listens on port `9113` and metrics about web traffic to the pod can be scraped there.
+1. [<container name>_pebble_ready](https://juju.is/docs/sdk/container-name-pebble-ready-event): fired on Kubernetes charms when the requested container is ready. Action: wait for the integrations, and configure the containers.
```
arturo-seijas, line 99: Some of the sentences after `Action:` are capitalized, some aren't

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+2. [config_changed](https://juju.is/docs/sdk/config-changed-event): usually fired in response to a configuration change using the GUI or CLI. Action: wait for the integrations, validate the configuration, update Ingress, and restart the containers.
+3. [leader_elected](https://juju.is/docs/sdk/leader-elected-event): is emitted for a unit that is elected as leader. Action: guarantee that all Indico workers have the same [secret key](https://docs.getindico.io/en/latest/config/settings/?highlight=secret_key#SECRET_KEY) that is used to sign tokens in URLs.
+4. [refresh_external_resources_action](https://charmhub.io/indico/actions): fired when refresh-external-resources action is requested. Action: Pull changes from the customization repository, reload uWSGI and upgrade the external plugins.
```
arturo-seijas, line 102: I'd go with `executed` instead of `requested`

docs/explanation/charm-architecture.md
amandahla: I'll add something small to give an overview so.

docs/explanation/charm-architecture.md
amandahla: Sorry, still working, will be available in my next commit.

docs/explanation/charm-architecture.md
arturo-seijas: I've added something for discourse, maybe you can leverage it :) https://github.com/canonical/discourse-k8s-operator/pull/21/files#diff-a91dcce1a3887d78aa883261a42dead6b363d5d3bc06e3b8838adbb16719ae19

docs/explanation/charm-architecture.md
amandahla: Done

docs/explanation/charm-architecture.md
amandahla: I didn't found but just in case change the Action to a new line, to be more visible.

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,103 @@
...
+5. [master_changed](https://github.com/canonical/ops-lib-pgsql): PostgreSQLClient custom event for when the connection details to the master database on this relation change. Action: Update the database connection string configuration and emit config_changed event.
```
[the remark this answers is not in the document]
amandahla: Added.

docs/explanation/charm-architecture.md
amandahla: Good catch

docs/explanation/charm-architecture.md
amandahla: Thanks!

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,148 @@
...
+indico-0                         6/6     Running   0         6h4m
 ```
-This shows there are 5 containers - the four named above, as well as a container for the charm code itself.
 
-Let's look at each container in turn in a bit more detail.
+This shows there are 6 containers - the five named above, as well as a container for the charm code itself.
+
+And if you run `kubectl describe pod indico-0`, all the containers will have as Command ```/charm/bin/pebble```. That's because Pebble is responsible for the processes startup as explained above.
+
+## Code Overview
```
arturo-seijas, line 28: Why is this section empty?

docs/explanation/charm-architecture.md
```diff
@@ -1,45 +1,148 @@
...
+
+## Charm Code Overview
+
+The `src/charm.py` is the default entry point for a charm and has the IndicoOperatorCharm Python class which inherits from CharmBase.
+
+CharmBase is the base class from which all Charms are formed, defined by [Ops](https://juju.is/docs/sdk/ops) (Python framework for developing charms).
+
+See more information in [Charm](https://juju.is/docs/sdk/constructs#heading--charm).
+
+The `__init__` method guarantees that the charm observes all events relevant to its operation and handles them.
```
arturo-seijas, line 135: The `__init__` method guarantees that the charm observes and handles all events relevant to its operation.

seems simpler to me

arturo-seijas (commented):
Are you planning to add the workflow changes to publish the code?

amandahla:
> Are you planning to add the workflow changes to publish the code?

There is a draft-publish-docs action already set in test_and_publish_charm. Would that be enough?

docs/explanation/charm-architecture.md
amandahla: (I answered that and then it counted as a review so I deleted it, anyway...)
I think would "overlap" the Juju Doc so keeped it only with the necessary to understand (in a very general way) what the charm does.

arturo-seijas (approved):
LGTM

arturo-seijas:
> > Are you planning to add the workflow changes to publish the code?
> 
> There is a draft-publish-docs action already set in test_and_publish_charm. Would that be enough?

Yes, sorry. My mistake

weiiwang01 (approved):

PR #198, opened 2022-12-12
feat(ISD-266): Collect metrics from Celery

amandahla:
* Add new container celery-prometheus-exporter (Merged https://github.com/danihodovic/celery-exporter/pull/186)

* Update unit and integration tests

* Add Grafana dashboard celery.json

* Update Indico Celery run command and Indico configuration file

* Update Explanation doc


files/etc/indico/indico.conf
```diff
@@ -5,6 +5,7 @@ ATTACHMENT_STORAGE = os.environ.get('ATTACHMENT_STORAGE', 'default')
 AUTH_PROVIDERS = ast.literal_eval(os.environ.get('INDICO_AUTH_PROVIDERS', 'None')) or {}
 BASE_URL = "{SERVICE_SCHEME}://{SERVICE_HOSTNAME}".format(**os.environ)
 CACHE_DIR        = "/srv/indico/cache"
+CELERY_CONFIG = ast.literal_eval(os.environ.get('CELERY_CONFIG', 'None')) or {}
```
jdkandersson, line 8: Out of curiosity, what is the reason for using en eval here? I see it above as well in existing code

src/charm.py
```diff
@@ -339,6 +343,38 @@ def _get_indico_nginx_pebble_config(self, _) -> Dict:
             },
         }
 
+    def _get_celery_prometheus_exporter_pebble_config(self, container) -> Dict:
```
jdkandersson, line 346: I couldn't spot use of the `container` argument?

src/charm.py
```diff
@@ -394,6 +430,24 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
...
 
+    def _get_celery_backed(self) -> str:
+        """Generate Celery Backend URL formed by Redis broker host and port.
+
+        Returns:
+            Celery Backend URL as expected by Indico and Celery Prometheus Exporter.
+        """
+        broker_rel = next(
+            rel
+            for rel in self.model.relations["redis"]
```
jdkandersson, line 441: from memory this code is repeated many times in this file, should it be some sort of helper function?

src/charm.py
```diff
@@ -394,6 +430,24 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
...
+            if rel.app and rel.app.name.startswith("redis-broker")
+        )
+        broker_unit = next(
+            unit for unit in broker_rel.data if unit.name.startswith("redis-broker")
```
jdkandersson, line 445: Similar comment as above

src/charm.py
```diff
@@ -459,6 +503,10 @@ def _get_indico_env_config(self, container: Container) -> Dict:
             },
         }
 
+        # Required for monitoring Celery
+        celery_config = {"worker_send_task_events": True, "task_send_sent_event": True}
+        env_config["CELERY_CONFIG"] = str(celery_config)
```
jdkandersson, line 508: Would it be better to use JSON here?

tests/integration/test_charm.py
```diff
@@ -49,7 +49,8 @@ async def test_prom_exporters_are_up(app: Application):
     """
     # Application actually does have units
     indico_unit = app.units[0]  # type: ignore
-    prometheus_targets = ["localhost:9113", "localhost:9102"]
+    # 9113 for NGINX, 9102 for StatsD and 9808 for Celery Prometheus exporters
+    prometheus_targets = ["localhost:9113", "localhost:9102", "localhost:9808"]
```
jdkandersson, line 53: Would an integration test with the new container be worth it?

tests/unit/test_charm.py
```diff
@@ -64,6 +64,8 @@ def test_all_pebble_services_ready(self):
         self.harness.set_leader(True)
 
         with patch.object(Container, "exec", return_value=MockExecProcess()):
+            self.harness.container_pebble_ready("celery-prometheus-exporter")
```
jdkandersson, line 67: Something to consider, probably not for this PR, should we convert these tests to pytest sometime soon?

docs/explanation/charm-architecture.md
```diff
@@ -13,15 +13,16 @@ Pebble `services` are configured through [layers](https://github.com/canonical/p
...
 As a result, if you run a `kubectl get pods` on a namespace named for the Juju model you've deployed the Indico charm into, you'll see something like the following:
 
 ```bash
 NAME                             READY   STATUS    RESTARTS   AGE
-indico-0                         6/6     Running   0         6h4m
+indico-0                         7/7     Running   0         6h4m
 ```
 
-This shows there are 6 containers - the five named above, as well as a container for the charm code itself.
+This shows there are 7 containers - the five named above, as well as a container for the charm code itself.
```
arturo-seijas, line 25: The six named above

files/etc/indico/indico.conf
arturo-seijas: Mapping the value of the environment variable, which is a string, to a Python object. Indico expects Python objects for some of those configurations.

src/charm.py
```diff
@@ -394,6 +430,24 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
             },
         }
 
+    def _get_celery_backed(self) -> str:
```
gregory-schiano, line 433: `backend` and not `backed`

src/charm.py
```diff
@@ -394,6 +430,24 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
...
+        broker_rel = next(
+            rel
+            for rel in self.model.relations["redis"]
+            if rel.app and rel.app.name.startswith("redis-broker")
+        )
+        broker_unit = next(
+            unit for unit in broker_rel.data if unit.name.startswith("redis-broker")
+        )
+        broker_host = broker_rel.data[broker_unit].get("hostname")
+        broker_port = broker_rel.data[broker_unit].get("port")
```
gregory-schiano, line 448: What happens if the relation is not yet available when the pebble ready event for this container is raised ?

src/charm.py
amandahla: Removed.

src/charm.py
amandahla: on_pebble_ready and on_config_changed wait for the relations by checking with are_relations_ready method.

tests/integration/test_charm.py
amandahla: Since is has no features besides providing a /metrics endpoint, this test should be enough.

src/charm.py
amandahla: Could you give an example?

src/charm.py
amandahla: Extracted _get_redis_cache_rel and _get_redis_broker_rel

src/charm.py
amandahla: Extracted _get_redis_cache_rel and _get_redis_broker_rel

src/charm.py
jdkandersson: `env_config["CELERY_CONFIG"] = json.dumps(celery_config)`

files/etc/indico/indico.conf
jdkandersson: Oh that is quite dangerous design on Indico's part, I guess we have to deal with it. We should consider the security implications of this

src/charm.py
```diff
@@ -430,17 +431,43 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
...
+        return next(
+            (
+                rel
+                for rel in self.model.relations["redis"]
+                if rel.app and rel.app.name.startswith("redis-broker")
+            ),
+            None,
+        )
+
+    def _get_redis_cache_rel(self) -> Relation:
```
jdkandersson, line 449: Is the type hint correct here? The default for `next` is `None` which could be returned. This could also be a logic problem for the cases where this function is called, do they correctly handle the `None` case?

src/charm.py
jdkandersson: Although I realise this is related to the earlier comment, I guess Indico expects this to be done in this way

merkata (dismissed):
Looks great!

src/charm.py
arturo-seijas: I think so. yes

src/grafana_dashboards/celery.json
```diff
@@ -0,0 +1,563 @@
...
+        "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [],
+  "annotations": {
+    "list": []
+  },
+  "description": "A dashboard that monitors Celery. It is created using the Celery-mixin for the the (Celery-exporter)[https://github.com/danihodovic/celery-exporter]",
+  "editable": true,
+  "gnetId": 17508,
```
arturo-seijas, line 18: Can you set this to null? iirc it will be overwritten anyway on import

src/grafana_dashboards/celery.json
```diff
@@ -0,0 +1,563 @@
...
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        }
+      ]
+    }
+  ],
+  "refresh": "",
```
arturo-seijas, line 498: A fresh value might come in handy. The other dashboard is using `5s`

files/etc/indico/indico.conf
arturo-seijas: In all the cases where this function is used to parse the env var, the env var value is not derived from user input, so we're safe here.

src/charm.py
amandahla: Type hint changed to Optional[Relation].
_are_relations_ready already did and now_get_celery_backend and _get_indico_env_config do too.

arturo-seijas (dismissed):
LGTM

docs/explanation/charm-architecture.md
```diff
@@ -59,15 +60,30 @@ The `NGINX Prometheus Exporter` is started with `-nginx.scrape-uri=http://localh
...
 - `9125`: UDP address on which to receive statsd metric.
 - `9102`: expose the web interface and generated Prometheus metrics. The metrics can be scraped by Prometheus here.
 
+### Celery Prometheus Exporter
+
+This container runs the `danihodovic/celery-exporter` image.
+
+The `Celery Prometheus Exporter` is started with:
+
+- `--broker-url=`: scrap metrics from the Redis (broker) container.
```
mthaddon, line 78: "scrape"?

arturo-seijas (dismissed):
LGTM

jdkandersson (dismissed):

merkata (dismissed):

src/charm.py
```diff
@@ -394,6 +431,52 @@ def _get_statsd_prometheus_exporter_pebble_config(self, _) -> Dict:
...
+            Relation between indico and redis-cache. If not found, returns None.
+        """
+        return next(
+            (
+                rel
+                for rel in self.model.relations["redis"]
+                if rel.app and rel.app.name.startswith("redis-cache")
+            ),
+            None,
+        )
```
gregory-schiano, line 462: Can't we factor even more these 2 methods ? These are exactly the same logic unless the name of the app we're looking for
I'd make 3 methods
2 that will call the 3rd method with "redis-cache" and "redis-broker" as extra parameter
And the 3rd method will take that extra parameter (str) and pass it to startswith

tests/integration/test_charm.py
gregory-schiano, line 53: Shouldn't we use constants for the port number ? They here and in charm.py

arturo-seijas (approved):
LGTM

merkata (approved):

PR #208, opened 2023-01-06
Load tests

arturo-seijas:
Runs Indico load tests every Sunday at 12

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -8,4 +8,7 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@load-tests
     secrets: inherit
     with:
+      chaos-enabled: true
+      chaos-experiments: pod-delete
+      chaos-app-label: app.kubernetes.io/name=indico
       load-test-enabled: true
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -7,3 +7,7 @@ jobs:
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
+    with:
+      chaos-enabled: true
+      chaos-experiments: pod-delete
+      chaos-app-label: app.kubernetes.io/name=indico
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -8,5 +8,4 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@load-tests
     secrets: inherit
     with:
-      chaos-enabled: false
       load-test-enabled: true
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -5,5 +5,7 @@ on:
 
 jobs:
   integration-tests:
-    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@load-tests
+    with:
+      load-test-enabled: true
     secrets: inherit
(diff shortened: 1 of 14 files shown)


gtrkiller (dismissed):
LGTM

merkata (commented):
Looks really good so far, solid foundation! :)

.github/workflows/load_test.yaml
```diff
@@ -0,0 +1,13 @@
+name: Load tests
+
+on:
+  schedule:
+    - cron: "0 12 * * 0"
```
nrobinaubertin, line 5: I'm not sure to understand the intent, could you add a short explanation for the scheduling ?

tests/conftest.py
```diff
@@ -1,7 +1,10 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+"""Fixtures for WordPress charm tests."""
```
nrobinaubertin, line 4: s/Wordpress/Indico-operator

nrobinaubertin (changes requested):

.github/workflows/integration_test.yaml
```diff
@@ -5,9 +5,10 @@ on:
 
 jobs:
   integration-tests:
-    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@load-tests
```
weiiwang01, line 8: I think it can wait until the `load-tests` branch merged into the `main` branch.

.github/workflows/load_test.yaml
arturo-seijas: Does this suffice?

tests/integration/conftest.py
```diff
@@ -1,6 +1,10 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+"""Fixtures for Indico charm integration tests."""
+
+# pylint:disable=redefined-outer-name
```
weiiwang01, line 6: This can be avoided by naming the fixture function to `*_fixture` and using the `name` argument in fixture decorator.

```python3
@fixture(scope="module", name="metadata")
def metadata_fixture():
    """Provides charm metadata."""
    yield yaml.safe_load(Path("./metadata.yaml").read_text("utf-8"))
```

arturo-seijas: Thanks

.github/workflows/load_test.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  load-tests:
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    with:
+      load-test-enabled: true
+      load-test-run-args: "-e LOAD_TEST_HOST=localhost"
```
amandahla, line 12: Would it be worth bootstrapping Indico before running the test?

arturo-seijas: This is just a dummy test but yes, this should be done when adding the proper tests

.github/workflows/load_test.yaml
nrobinaubertin: If we want to avoid running a workflow, is it a possibility to avoid running it on commit that have a keyword (like `wf:no-load-test`) ?

arturo-seijas: Why would be want a mechanism to bypass the test?

nrobinaubertin: If we are changing a minor thing, or something that would not impact load tests, this would avoid the waiting time (and spare some resources). But I agree that it's out of scope and optional

arturo-seijas: It feels like over optimization. The compute resources are free (at least for now)

nrobinaubertin (approved):

gtrkiller (approved):
LGTM

weiiwang01 (approved):

PR #210, opened 2023-01-12
test(ISD-374): Add OWASP Zap test

amandahla:
For reviewers:
The ZAP action was added at the Trivy PR so this one ended up just changing minor things.

https://github.com/canonical/indico-operator/pull/216


arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, *args):
             redis_relation={},
         )
 
-        self.db = pgsql.PostgreSQLClient(self, "db")
+        self.db = pgsql.PostgreSQLClient(self, "db")  # pylint: disable=C0103
```
jdkandersson, line 79: What is the reason for the disable here please?

tests/integration/conftest.py
```diff
@@ -49,7 +53,7 @@ async def app(
     nginx_prometheus_exporter_image: str,
     statsd_prometheus_exporter_image: str,
     celery_prometheus_exporter_image: str,
-):
+):  # pylint:disable=R0913
```
jdkandersson, line 56: What is the reason for disable this here please?

tests/zap/add_header_request.py
```diff
@@ -0,0 +1,11 @@
+# pylint: skip-file
+headers = dict({"Host": "indico.local"})
+
+
+def sendingRequest(msg, initiator, helper):  # noqa: N802
+    for h in list(headers):
```
jdkandersson, line 6: Should we call the variable `header`? `h` could mean a lot of things

tests/zap/add_header_request.py
```diff
@@ -0,0 +1,11 @@
+# pylint: skip-file
+headers = dict({"Host": "indico.local"})
```
jdkandersson, line 2: Would it be worth making this a tuple with 2 element tuples for each header name and value? and then it can be a constant? And the arguments to `.setHeader` would just be a `*` on the tuple with the header

tests/zap/add_header_request.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def sendingRequest(msg, initiator, helper):  # noqa: N802
```
jdkandersson, line 5: What is the reason for this naming convention please?

tests/zap/hook.py
```diff
@@ -0,0 +1,19 @@
+# pylint: skip-file
+def zap_started(zap, target):
+    print(
```
jdkandersson, line 3: Should we use `logging` here? Similar question for the remaining `print` statements

tox.ini
```diff
@@ -92,10 +101,10 @@ commands =
...
-    pytest
     juju
-    pytest-operator
+    pytest
     pytest-asyncio
+    pytest-operator
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
```
jdkandersson, line 110: Final new line?

jdkandersson (commented):
Thanks for adding all that documentation!

tests/integration/conftest.py
amandahla: "R0913: Too many arguments (6/5) (too-many-arguments)"
I didn't test but there is an option to change that: "max-args".

tests/zap/add_header_request.py
amandahla: To be honest, I wrote like that because of this [example](https://github.com/zaproxy/community-scripts/blob/main/httpsender/add_header_request.py).

Now that you mentioned it and since it's only one header, we could just set it like this:
"msg.getRequestHeader().setHeader("Host", "indico.local)"

What do you think?

.github/workflows/integration_test.yaml
```diff
@@ -11,3 +11,7 @@ jobs:
       chaos-enabled: true
       chaos-experiments: pod-delete
       chaos-app-label: app.kubernetes.io/name=indico
+      zap-enabled: true
+      zap-target-port: 8080
+      zap-before-command: "curl -H \"Host: indico.local\" $ZAP_TARGET:8080/bootstrap --data-raw 'csrf_token=00000000-0000-0000-0000-000000000000&first_name=admin&last_name=admin&email=admin%40admin.com&username=admin&password=lunarlobster&confirm_password=lunarlobster&affiliation=Canonical'"
```
gregory-schiano, line 16: Order it alphabeticaly ?

tests/integration/conftest.py
gregory-schiano: Could it be interesting to make a fixture that will return the 3 images in a dict ?

tests/zap/add_header_request.py
amandahla: This is defined and expected by ZAP tool. The template:
https://github.com/zaproxy/zap-extensions/blob/main/addOns/jython/src/main/zapHomeFiles/scripts/templates/httpsender/HttpSender%20default%20template.py

tests/zap/add_header_request.py
amandahla: Removed the for as suggested in the next answer.

arturo-seijas (changes requested):
There're lots of changes related to linting issues that should be solved already in main. Can you merge/rebase?

tests/conftest.py
```diff
@@ -1,7 +1,10 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+"""Fixtures for WordPress charm tests."""
```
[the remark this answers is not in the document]
amandahla: Merged.

tests/integration/conftest.py
amandahla: Good idea!

merkata:
I don't know why Github shows all checks as passed, that's confusing, the actual run is in the integration tests and they are currently failing:

```text
[Invalid workflow file: .github/workflows/integration_test.yaml#L17](https://github.com/canonical/indico-operator/actions/runs/3932624507/workflow)
The workflow is not valid. .github/workflows/integration_test.yaml (Line: 17, Col: 20): Invalid input, zap-options is not defined in the referenced workflow.
```
Can you have a look @amandahla?

amandahla:
> I don't know why Github shows all checks as passed, that's confusing, the actual run is in the integration tests and they are currently failing:
> 
> ```
> [Invalid workflow file: .github/workflows/integration_test.yaml#L17](https://github.com/canonical/indico-operator/actions/runs/3932624507/workflow)
> The workflow is not valid. .github/workflows/integration_test.yaml (Line: 17, Col: 20): Invalid input, zap-options is not defined in the referenced workflow.
> ```
> 
> Can you have a look @amandahla?

Thanks for pointing that out! Fixed.

src/charm.py
jdkandersson: I think `pylint` lets you configure `# pylint: disable=invalid-name` which means it is easier to know what the issue is. Would it be worth adding a comment explaining why this rule should be disabled?

tests/zap/add_header_request.py
jdkandersson: Sounds reasonable too

tests/integration/conftest.py
```diff
@@ -27,33 +26,30 @@ def app_name_fixture(metadata):
...
-
-
-@fixture(scope="module", name="celery_prometheus_exporter_image")
-def celery_prometheus_exporter_image_fixture(metadata):
-    """Provides the celery prometheus exporter image from the metadata."""
-    yield metadata["resources"]["celery-prometheus-exporter-image"]["upstream-source"]
+@fixture(scope="module", name="prometheus_exporter_images")
+def prometheus_exporter_images_fixture(metadata):
+    """Provides Prometheus exporter images from the metadata."""
+    prometheus_exporter_images = {
```
jdkandersson, line 32: could make this a `typing.NamedTuple` for easier auto-complete and to make sure that the dictionary can't be accidentally modified by other tests, what do you think?

.github/workflows/integration_test.yaml
```diff
@@ -8,7 +8,11 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
     with:
+      chaos-app-label: app.kubernetes.io/name=indico
       chaos-enabled: true
       chaos-experiments: pod-delete
-      chaos-app-label: app.kubernetes.io/name=indico
       load-test-enabled: true
+      zap-before-command: "curl -H \"Host: indico.local\" $ZAP_TARGET:8080/bootstrap --data-raw 'csrf_token=00000000-0000-0000-0000-000000000000&first_name=admin&last_name=admin&email=admin%40admin.com&username=admin&password=lunarlobster&confirm_password=lunarlobster&affiliation=Canonical'"
```
arturo-seijas, line 15: Can you execute this against the ingress instead? It should be available in localhost:80

.github/workflows/integration_test.yaml
```diff
@@ -8,7 +8,11 @@ jobs:
...
+      zap-enabled: true
+      zap-cmd-options: '-T 60 -z "-addoninstall jython" --hook "/zap/wrk/tests/zap/hook.py"'
+      zap-target-port: 8080
```
arturo-seijas, line 18: Same here, I guess

tests/integration/conftest.py
amandahla: It's a good point but I left it as a dict because the only place that uses this is in the contest file itself in the app method as keys to "resources" dict. So returning a dict is an easier way to update this var.

What do you think? Could you please give an example of how it would work with NamedTuple?


```
    resources = {
        "indico-image": pytestconfig.getoption("--indico-image"),
        "indico-nginx-image": pytestconfig.getoption("--indico-nginx-image"),
    }
    resources.update(prometheus_exporter_images)
```

.github/workflows/integration_test.yaml
amandahla: Sure, fixed.

arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -5,14 +5,16 @@ on:
 
 jobs:
   integration-tests:
-    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@ISD-374-owasp-baseline
```
jdkandersson, line 8: I'm guessing we won't merge until we can point this back to main?

tests/integration/conftest.py
jdkandersson: It is probably ok on this case since it is configuration passed in as a dictionary

tests/zap/add_header_request.py
```diff
@@ -1,7 +1,27 @@
-# pylint: skip-file
+# pylint: disable=missing-module-docstring,invalid-name,unused-argument
```
jdkandersson, line 1: The new guidance on disabling linting rules includes adding a comment above the disable line explaining why the rule is bing disabled, is that applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

tests/zap/hook.py
```diff
@@ -1,8 +1,14 @@
-# pylint: skip-file
+# pylint: disable=missing-module-docstring,invalid-name,unused-argument
```
jdkandersson, line 1: The new guidance on disabling linting rules includes adding a comment above the disable line explaining why the rule is bing disabled, is that applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

tests/zap/hook.py
```diff
@@ -1,8 +1,14 @@
...
 import logging
 
 
 def zap_started(zap, target):
+    """Actions when starts
+
+    Args:
+        zap (ZAPv2): ZAPv2 instance
+        target (string): Target being scanned
```
jdkandersson, line 10: Would it be worth putting the type hints in the function signature instead?

arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -5,13 +5,16 @@ on:
...
     secrets: inherit
     with:
+      chaos-app-label: app.kubernetes.io/name=indico
       chaos-enabled: true
       chaos-experiments: pod-delete
-      chaos-app-label: app.kubernetes.io/name=indico
+      load-test-enabled: true
+      load-test-run-args: "-e LOAD_TEST_HOST=localhost"
+      zap-before-command: "curl -H \"Host: indico.local\" http://localhost/bootstrap --data-raw 'csrf_token=00000000-0000-0000-0000-000000000000&first_name=admin&last_name=admin&email=admin%40admin.com&username=admin&password=lunarlobster&confirm_password=lunarlobster&affiliation=Canonical'"
```
gregory-schiano, line 16: I think Niels is doing something to use the CLI in non-interactive shell, could be interesting to sync with him and align the approach for the boostrap

amandahla: @nrobinaubertin Is it possible to bootstrap Indico using CLI?

.github/workflows/integration_test.yaml
amandahla: Yes, I just did this change to confirm that the baseline scan would work. :)

.github/workflows/integration_test.yaml
nrobinaubertin: Right now, I'm working on patching the indico CLI to be able to add users non-interactively. Adding a user will dismiss the bootstrapping form.  
Since I'm patching the CLI and using a custom container for that, I think Amanda's solution is best for this PR

amandahla: Thanks!

tests/zap/hook.py
amandahla: Pylance complains that ZAPv2 is not defined.

jdkandersson: Maybe it needs to be imported? Type hints can also be strings?

arturo-seijas (dismissed):
Nice!

tests/zap/hook.py
amandahla: I didn't want to import ZAPv2 because it's a hook script and I didn't find a way to add the Type Hint without importing it :-(
For now, I'll leave it just as a docstring but if you have an example of how to do it, that would be great.

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #211, opened 2023-01-13
Use secrets in peer relation

arturo-seijas:
Store the secret-key used to coordinate the Indico instances leveraging the new Juju secrets feature when supported


src/charm.py
```diff
@@ -522,6 +523,15 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
         peer_relation = self.model.get_relation("indico-peers")
 
+        secret_value = None
+        if peer_relation and not self._has_secrets():
+            secret_value = peer_relation.data[self.app].get("secret-key")
+        elif peer_relation and self._has_secrets():
+            secret_id = peer_relation.data[self.app].get("secret-id")
+            if secret_id:
+                secret = self.model.get_secret(id=secret_id)
+                secret_value = secret.get_content().get("secret-key")
```
mthaddon, line 533: Feels like it might be worth breaking this out to a separate function?

src/charm.py
```diff
@@ -497,6 +497,23 @@ def _get_celery_backend(self) -> str:
...
+        """Return the Indico secret key needed to deploy multiple Indico instances.
+
+        Returns:
+            Indico secret key.
+        """
+        secret_value = None
+        peer_relation = self.model.get_relation("indico-peers")
+        if peer_relation and not self._has_secrets():
+            secret_value = peer_relation.data[self.app].get("secret-key")
+        elif peer_relation and self._has_secrets():
```
gregory-schiano, line 510: May I suggest:
```
if not peer_relation:
   return None

if self._has_secrets():
...
else
...

return secret_value
```

The goal is to avoid `if` `elif` without `else` which I try to always avoid

arturo-seijas: Any reason why?

gregory-schiano: Few reasons which are linked together:
- Readability
- It's often part of best practices and enforced in static code analysis rules
- It often shows a logic non completness
- It's part of defensive programming rules

arturo-seijas: Done. Thx for the explanation

mthaddon (dismissed):
Good to land now that ops 2.0.0 is out!

gregory-schiano (dismissed):
LGTM, let's go with our first Juju secret :)

gregory-schiano (approved):
You'll make it Arturo !

mthaddon (approved):

PR #219, opened 2023-01-22
Add helpers functions to create the model

nrobinaubertin:
The goal is to be able to reuse the same model when replaying the integration tests.  
This can also be helpful to split future integrations tests in multiple files (as the operator-workflows code keeps the model between different integration test modules).

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 """Helper functions for integration tests.
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Helper functions for integration tests."""
+
+from typing import Optional
+
+from ops.model import Application
+from pytest_operator.plugin import OpsTest
+
+
+async def deploy_related_charm(
+    ops_test: OpsTest,
+    charm_name: str,
+    instance_name: Optional[str] = None,
+    raise_on_error: bool = True,
+    trust: bool = False,
+) -> Application:
+    """Deploy a charm related to our main app, if it's not already deployed."""
+    assert ops_test.model
+    instance_name = instance_name or charm_name
+    if instance_name not in ops_test.model.applications:
+        instance = await ops_test.model.deploy(charm_name, instance_name, trust=trust)
+        await ops_test.model.wait_for_idle(raise_on_error=raise_on_error)
+        return instance
+    return ops_test.model.applications[instance_name]
+
+
+# pylint: disable=dangerous-default-value
+async def deploy_app(
+    ops_test: OpsTest,
+    app_name: str,
+    series: str,
+    resources: dict = {},
+) -> Application:
+    """Deploy our main app, if it's not already deployed."""
+    assert ops_test.model
+    if app_name not in ops_test.model.applications:
+        charm = await ops_test.build_charm(".")
+        application = await ops_test.model.deploy(
+            charm, resources=resources, application_name=app_name, series=series
+        )
+        await ops_test.model.wait_for_idle()
+        return application
+    return ops_test.model.applications[app_name]
+
+
+async def app_add_relation(
+    ops_test: OpsTest,
+    app_name: str,
+    relation_id: str,
+):
+    """Relate to applications if the relation does not already exists."""
+    assert ops_test.model
+    assert ops_test.model.relations
+
+    if not any(filter(lambda x: x.matches(relation_id), ops_test.model.relations)):  # type: ignore
+        await ops_test.model.add_relation(app_name, relation_id)
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -2,7 +2,10 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Helper functions for integration tests."""
+"""Helper functions for integration tests.
+
+These helper functions will deploy charms or add relations only if they do not already exists.
+"""
 
 from typing import Optional
 
@@ -27,15 +30,16 @@ async def deploy_related_charm(
     return ops_test.model.applications[instance_name]
 
 
-# pylint: disable=dangerous-default-value
 async def deploy_app(
     ops_test: OpsTest,
     app_name: str,
     series: str,
-    resources: dict = {},
+    resources: Optional[dict] = None,
 ) -> Application:
     """Deploy our main app, if it's not already deployed."""
     assert ops_test.model
+    if resources is None:
+        resources = {}
     if app_name not in ops_test.model.applications:
         charm = await ops_test.build_charm(".")
         application = await ops_test.model.deploy(
@@ -46,7 +50,7 @@ async def deploy_app(
     return ops_test.model.applications[app_name]
 
 
-async def app_add_relation(
+async def relate_app(
     ops_test: OpsTest,
     app_name: str,
     relation_id: str,
@@ -55,5 +59,5 @@ async def app_add_relation(
     assert ops_test.model
     assert ops_test.model.relations
 
-    if not any(filter(lambda x: x.matches(relation_id), ops_test.model.relations)):  # type: ignore
+    if not any(relation for relation in ops_test.model.relations if relation.matches(relation_id)):
         await ops_test.model.add_relation(app_name, relation_id)
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -80,7 +80,7 @@ async def relate_app(
 
     Args:
         ops_test: The pytest operator plugin instance.
-        app_name: The application name that will be part of the realtion.
+        app_name: The application name that will be part of the relation.
         relation_id: The relation identifier.
 
     Returns:
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -20,7 +20,18 @@ async def deploy_related_charm(
     raise_on_error: bool = True,
     trust: bool = False,
 ) -> Application:
-    """Deploy a charm related to our main app, if it's not already deployed."""
+    """Deploy a charm related to our main app, if it's not already deployed.
+
+    Args:
+        ops_test: The pytest operator plugin instance.
+        charm_name: The name of the charm to deploy.
+        instance_name: The name that the instance will have.
+        raise_on_error: Should an Exception be raised if the charm enters an error state.
+        trust: Should we trust the charm to change things on the k8s substrate.
+
+    Returns:
+        The deployed application.
+    """
     assert ops_test.model
     instance_name = instance_name or charm_name
     if instance_name not in ops_test.model.applications:
@@ -36,7 +47,17 @@ async def deploy_app(
     series: str,
     resources: Optional[dict] = None,
 ) -> Application:
-    """Deploy our main app, if it's not already deployed."""
+    """Deploy our main app, if it's not already deployed.
+
+    Args:
+        ops_test: The pytest operator plugin instance.
+        app_name: The name of the application's charm.
+        series: The Ubuntu series with which to deploy the charm.
+        resources: A dict with resources (like container images).
+
+    Returns:
+        The deployed application.
+    """
     assert ops_test.model
     if resources is None:
         resources = {}
@@ -55,7 +76,16 @@ async def relate_app(
     app_name: str,
     relation_id: str,
 ):
-    """Relate to applications if the relation does not already exists."""
+    """Relate to applications if the relation does not already exists.
+
+    Args:
+        ops_test: The pytest operator plugin instance.
+        app_name: The application name that will be part of the realtion.
+        relation_id: The relation identifier.
+
+    Returns:
+        The created relation.
+    """
     assert ops_test.model
     assert ops_test.model.relations
 
--- a/tox.ini
+++ b/tox.ini
@@ -36,12 +36,12 @@ description = Check code against coding style standards
 deps =
     black
     codespell
-    flake8<6.0.0
     flake8-builtins
     flake8-copyright<6.0.0
-    flake8-docstrings>=1.6.0
     flake8-docstrings-complete>=1.0.3
+    flake8-docstrings>=1.6.0
     flake8-test-docs>=1.0
+    flake8<6.0.0
     isort
     mypy
     pep8-naming
--- a/tox.ini
+++ b/tox.ini
@@ -36,12 +36,12 @@ description = Check code against coding style standards
 deps =
     black
     codespell
+    flake8<6.0.0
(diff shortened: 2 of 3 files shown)


tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Helper functions for integration tests."""
```
merkata, line 5: I like this - can you expand the docstring to include the current function and how they help with integration tests, would be useful imo. Thanks!

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    assert ops_test.model
+    instance_name = instance_name or charm_name
+    if instance_name not in ops_test.model.applications:
+        instance = await ops_test.model.deploy(charm_name, instance_name, trust=trust)
+        await ops_test.model.wait_for_idle(raise_on_error=raise_on_error)
+        return instance
+    return ops_test.model.applications[instance_name]
+
+
+# pylint: disable=dangerous-default-value
```
merkata, line 30: Could you explain why the disable is necessary here, it sounds like a "famous last words" type of thing ("Let's just disable this dangerous default value here...").

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+        charm = await ops_test.build_charm(".")
+        application = await ops_test.model.deploy(
+            charm, resources=resources, application_name=app_name, series=series
+        )
+        await ops_test.model.wait_for_idle()
+        return application
+    return ops_test.model.applications[app_name]
+
+
+async def app_add_relation(
```
merkata, line 49: I see the other functions are <action>_<kind> as in deploy_app, but here it is the other way around - perhaps change it?

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    ops_test: OpsTest,
+    app_name: str,
+    relation_id: str,
+):
+    """Relate to applications if the relation does not already exists."""
+    assert ops_test.model
+    assert ops_test.model.relations
+
+    if not any(filter(lambda x: x.matches(relation_id), ops_test.model.relations)):  # type: ignore
```
merkata, line 58: Why is the ignore there, could this be refactored so we don't ignore it?

merkata:
I like this @nrobinaubertin ! Left some comments to kick off a discussion so I can better understand things. Looks really good.

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+
+"""Helper functions for integration tests."""
+
+from typing import Optional
+
+from ops.model import Application
+from pytest_operator.plugin import OpsTest
+
+
+async def deploy_related_charm(
```
jdkandersson, line 13: I don't follow the need for this kind of helper, hopefully you can help me out. When would we need to conditionally add a related charm based on whether it is already deployed? Wouldn't the test suit either deploy it or not?

The scenario I can think of, maybe there are others, are when only some tests require certain related charms. If that is the case, wouldn't it be better to add fixtures with those related charms and add those fixtures to the tests requiring those related charms to be deployed?

A similar question applies to `app_add_relation`

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    instance_name = instance_name or charm_name
+    if instance_name not in ops_test.model.applications:
+        instance = await ops_test.model.deploy(charm_name, instance_name, trust=trust)
+        await ops_test.model.wait_for_idle(raise_on_error=raise_on_error)
+        return instance
+    return ops_test.model.applications[instance_name]
+
+
+# pylint: disable=dangerous-default-value
+async def deploy_app(
```
jdkandersson, line 31: I don't follow this either for similar reason as `deploy_related_charm`, under what conditions would we not be sure whether the charm is already deployed?

tests/integration/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    ops_test: OpsTest,
+    app_name: str,
+    series: str,
+    resources: dict = {},
```
jdkandersson, line 35: Using dictionaries on default values is quite dangerous as `pylint` suggests, this instance will be shared among *all* calls to this function. For example, if `ops_test.model.deploy` makes any changes to the dictionary, which it might, those changes will persist for all future calls to `deploy_app`, regardless of which charm is being deployed

tests/integration/helpers.py
jdkandersson, line 58: Would it be worth expanding on the variable name `x`? I'm not sure what it is. This can also be refactored to: `any(relation for relation in ops_test.model.relations if relation.matches(relation_id))`, do you think that would be better?

jdkandersson (commented):
Are the helpers required? On surface level it appears that using fixtures might be a better alternative, perhaps I'm missing something?

tests/integration/helpers.py
nrobinaubertin: I found out when working on the k6 load tests that I couldn't add integ tests in another file because the model is kept between module calls (due to [this](https://github.com/canonical/operator-workflows/blob/main/.github/workflows/integration_test.yaml#L302)) and the current fixtures didn't take that into account.  

I can't change the scope of the fixtures to "package" because OpsTest himself is [declared as module](https://github.com/charmed-kubernetes/pytest-operator/blob/main/pytest_operator/plugin.py#L220).

I decided to create helper functions outside of the fixtures because I found it more clean and eventually reusable in other repositories if it's working well.

tests/integration/helpers.py
nrobinaubertin: Thanks, I didn't think of this structure

tests/integration/helpers.py
nrobinaubertin: Got it

tests/integration/helpers.py
nrobinaubertin: I used David's proposal for this one

tests/integration/helpers.py
nrobinaubertin: Changed the name to `relate_app()`

tests/integration/helpers.py
nrobinaubertin: Fixed

tests/integration/helpers.py
nrobinaubertin: I added an explanation line. What do you think ?

tests/integration/helpers.py
```diff
@@ -0,0 +1,63 @@
...
+    ops_test: OpsTest,
+    charm_name: str,
+    instance_name: Optional[str] = None,
+    raise_on_error: bool = True,
+    trust: bool = False,
+) -> Application:
+    """Deploy a charm related to our main app, if it's not already deployed."""
+    assert ops_test.model
+    instance_name = instance_name or charm_name
+    if instance_name not in ops_test.model.applications:
```
gregory-schiano, line 26: Perhaps you should also check the state of the application ? Is idle ? Is it active ?

tests/integration/helpers.py
```diff
@@ -0,0 +1,63 @@
...
+async def deploy_app(
+    ops_test: OpsTest,
+    app_name: str,
+    series: str,
+    resources: Optional[dict] = None,
+) -> Application:
+    """Deploy our main app, if it's not already deployed."""
+    assert ops_test.model
+    if resources is None:
+        resources = {}
```
gregory-schiano, line 42: Couldn't you directly put a default value to `{}` ?

tests/integration/helpers.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
```
arturo-seijas, line 2: 2023

tox.ini
```diff
@@ -36,12 +36,12 @@ description = Check code against coding style standards
 deps =
     black
     codespell
-    flake8<6.0.0
     flake8-builtins
     flake8-copyright<6.0.0
-    flake8-docstrings>=1.6.0
     flake8-docstrings-complete>=1.0.3
+    flake8-docstrings>=1.6.0
```
arturo-seijas, line 42: Why this change?

tests/integration/helpers.py
```diff
@@ -0,0 +1,63 @@
...
+    instance_name: Optional[str] = None,
+    raise_on_error: bool = True,
+    trust: bool = False,
+) -> Application:
+    """Deploy a charm related to our main app, if it's not already deployed."""
+    assert ops_test.model
+    instance_name = instance_name or charm_name
+    if instance_name not in ops_test.model.applications:
+        instance = await ops_test.model.deploy(charm_name, instance_name, trust=trust)
+        await ops_test.model.wait_for_idle(raise_on_error=raise_on_error)
```
arturo-seijas, line 28: Won't this `await`effectively make impossible to deploy the charms in parallel?

arturo-seijas:
> The goal is to be able to reuse the same model when replaying the integration tests.
Why would you want to replay the tests? If you reuse the model there's a high chance it is keeping state between executions, why not redeploying from scratch?

nrobinaubertin:
> Why would you want to replay the tests? If you reuse the model there's a high chance it is keeping state between executions, why not redeploying from scratch?

Initially, I had no choice if I wanted to split the tests across multiple files due to the reasons explained [here](https://github.com/canonical/indico-operator/pull/219#discussion_r1083752037).  
But when developing, I find it convenient to just relaunch the test I want with tox without having to redeploy everything (when I know that it won't collide with what I done earlier).  
There's still the possibility to tear everything up beforehand.

tox.ini
nrobinaubertin: I just sorted the deps with `sort`. It's a choice to make (IMHO) future diffs more readable. Happy to revert if out-of-scope/no-wanted

tests/integration/helpers.py
nrobinaubertin: No because [pylint](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html) and [David](https://github.com/canonical/indico-operator/pull/219#discussion_r1083616100) will not be happy. And with good reasons

tox.ini
arturo-seijas: I was asking because I would have considered already sorted

tests/integration/helpers.py
jdkandersson: I think the problem you have might be that clean up doesn't happen properly after the tests at the module level are completed. One option is to wait for the cleanup to complete after the integration tests and for the re-deployment to occur before the load tests.

For other charms (I believe wordpress and Jenkins do this), when tests need a clean model in ci, you can use a matrix workflow by using the module parameter on the integration tests. That way you get a clean model for the k6 tests which is separate from the integration tests. You also get to run the tests in parellel, speeding up the test suit. E.g., add a `pytest.mark` as `integration` on the integration tests and `pytest.mark` of `load` for the load tests and then pass `["integration", "load"]` as the module parameter on the integration tests.

tests/integration/helpers.py
jdkandersson: In Python a mutable value should never be set as a default on a parameter because that default will be shared for all function calls that don't pass a value for that argument. This is because the default parameter is evaluated at function definition time, not at function call time. So anything like a dict, list, object etc should not be a default value. Strings, integers, tuples and other immutable types are safe because they can't be modified. 

If you use a mutable value and it gets modified in a function call, that modification persist to the *next* function call. It is a bit counter intuitive, hence the pylint warning

tests/integration/helpers.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+async def deploy_related_charm(
+    ops_test: OpsTest,
+    charm_name: str,
+    instance_name: Optional[str] = None,
+    raise_on_error: bool = True,
+    trust: bool = False,
+) -> Application:
+    """Deploy a charm related to our main app, if it's not already deployed."""
```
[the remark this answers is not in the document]
jdkandersson: Shouldn't this get enforced by the `flake8-docstrings-complete` linter?

tests/integration/helpers.py
jdkandersson: If you really want to avoid this you could come up with a decorator that takes an argument name and a class for the default value, inspects the kwargs of a function call and if a given value is `None`, replaces it with an instance of that class. This way you can hide the `None` check in the function. That decorator could be generalised so that it can be used in any case where a mutable default is required

jdkandersson: Here is a starting point from ChatGPT, it is still wrong I think because it will always replace the value rather than only if the default is detected, I would code it differently
Screenshot 2023-01-23 at 10 30 48 pm

jdkandersson: maybe an additional check on `not value` would do the trick as well

jdkandersson:
> > Why would you want to replay the tests? If you reuse the model there's a high chance it is keeping state between executions, why not redeploying from scratch?
> 
> Initially, I had no choice if I wanted to split the tests across multiple files due to the reasons explained [here](https://github.com/canonical/indico-operator/pull/219#discussion_r1083752037). But when developing, I find it convenient to just relaunch the test I want with tox without having to redeploy everything (when I know that it won't collide with what I done earlier). There's still the possibility to tear everything up beforehand.

Or just use a matrix and make the integration and k6 tests separate matrix items so they don't interfere with each other

tests/integration/helpers.py
gregory-schiano: Wow... didn't this for python... any reference on why this design choice oO
Coming from other languages this sounds like an ugly behavior (evaluating the default at function definition and not function call time)

tests/integration/helpers.py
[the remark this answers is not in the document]
gregory-schiano: Perhaps not in tests ?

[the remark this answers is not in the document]
nrobinaubertin: I added args an return descriptions

tests/integration/helpers.py
nrobinaubertin: Done

tox.ini
nrobinaubertin: I'm using neovim. I needed to use `:sort l` to sort with the current collation locale to have the same sort as you. I'll do that from now on.

tests/integration/helpers.py
jdkandersson: I think it is deliberate, not specifically intending to cause this problem, instead so that Python functions are first class objects. You can read this stack overflow if you would like to see some debate: https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

The summary is that it is a side-effect of making python functions first-class objects, you can think of these defaults as equivalent to class attributes, where it is also not advisable to default them to mutable objects - unless that is the intended behaviour, e.g., the object should maintain state across all instances. Probably not usually a good idea though.

tests/integration/helpers.py
nrobinaubertin: I see.  
Then the only added value of this PR would be the case where we want to run the full integ tests locally or on a CI that is not github. And even then, we would be better off having a "good" clean up.

nrobinaubertin:
I'm closing this in the light of David's comment: https://github.com/canonical/indico-operator/pull/219#discussion_r1083930067

PR #220, opened 2023-01-22
Add an add-admin action

nrobinaubertin:
The goal is to add a add-admin action to the charm that uses a modified version of indico CLI tool.  
I added a `click` command group (via plugin) to allow the `indico autocreate admin <email> <password>` command.


actions.yaml
```diff
@@ -1,2 +1,24 @@
...
+    first-name:
+      type: string
+      description: User firstname.
+    last-name:
+      type: string
+      description: User lastname.
+    affiliation:
+      type: string
+      description: User affiliation.
+    is-admin:
```
arturo-seijas, line 21: Generally, users will authenticated by local accounts, so I'd suggest to create only admin users to simplify. In fact, if possible I'd add only username and password as params.

nrobinaubertin: Even if it's just for testing purposes, isn't it better to have more options ?

src/charm.py
```diff
@@ -674,6 +677,12 @@ def _get_indico_env_config(self, container: Container) -> Dict:
             env_config = {**env_config, **self._get_http_proxy_configuration()}
         return env_config
 
+    def _get_indico_env_config_str(self, container: Container) -> Dict[str, str]:
```
nrobinaubertin, line 680: To pass the env to container.exec(), it must be `Dict[str, str]`. I'm not sure that the way I've done it (by simply taking the str() of each value) is valid.

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
+# This file is part of Indico.
```
jdkandersson, line 1: To help me understand, what is the reason for making this a separate cli rather than put it into the charm code? Generally speaking, we would probably prefer to have this in the charm source code (perhaps as a separate module). That way the logic is re-usable, the logic of the charm lives in the source code and is easily discoverable, it is clear to what degree it is covered by our testing, it is clear whether it is covered by our linting, and so on

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+from indico.util.date_time import utc_to_server
+from terminaltables import AsciiTable
+
+
+@cli_group()
+def cli():
+    pass
+
+
+def _print_user_info(user):
```
jdkandersson, line 28: Would this be worth a refactor to make a function that returns a string so that we can test it easily? And then perhaps print the output of that function?

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+    print(f"  Family name: {user.last_name}")
+    print(f"  Email: {user.email}")
+    print(f"  Affiliation: {user.affiliation}")
+    if flags:
+        print(cformat("  Flags: {}".format(", ".join(flags))))
+    print()
+
+
+def _safe_lower(s):
+    return (s or "").lower()
```
jdkandersson, line 50: What about `s.lower() if s else ""`?

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+
+
+def _format_dt(dt):
```
jdkandersson, line 53: Should this have a docstring? It doesn't seem clear what this function is intended to achieve?

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+    substring,
+    include_deleted,
+    include_pending,
+    include_blocked,
+    include_external,
+    include_system,
+    **criteria,
+):
+    """Search users matching some criteria."""
+    assert set(criteria.keys()) == {"first_name", "last_name", "email", "affiliation"}
```
jdkandersson, line 91: Will this produce a nice output for the user if something goes wrong? `criteria.keys()` already is equivalent to a set btw, passing it to the `set` constructor is not necessary

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+    criteria = {k: v for k, v in criteria.items() if v is not None}
```
jdkandersson, line 92: should the variable names be expanded here for readability? Generally, single letter variables (even in comprehensions) are difficult to read. expanding to `key` `value` etc doesn't add much length and is much easier to read. Similar comment for the other single character variables in this file

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+    res = search_users(
+        exact=(not substring),
+        include_deleted=include_deleted,
+        include_pending=include_pending,
+        include_blocked=include_blocked,
+        external=include_external,
+        allow_system_user=include_system,
+        **criteria,
+    )
+    if not res:
```
jdkandersson, line 102: Should we break this up into a few sub functions?

jdkandersson (commented):
Haven't finished this review yet, there are some broad questions about whether this should be a CLI or it should be in the charm source code. Let's address that first and then I'll finish the review

nrobinaubertin:
> Haven't finished this review yet, there are some broad questions about whether this should be a CLI or it should be in the charm source code. Let's address that first and then I'll finish the review

Sorry, it probably wasn't obvious from the PR description. I'm patching a file from the source code of indico ([here](https://github.com/indico/indico/blob/master/indico/cli/user.py) is the source).  
I decided to do it that way to avoid fiddling with their internal database structure.

So I want to do it in a way that is simple, ideally robust to some unrelated changes in indico and better, that breaks when there are changes that would affect this change.  
I thought of 3 ways to do it: 
1) wget the source and adding my bit at the end
2) wget the source and apply a git patch
3) replacing the whole file

I decided to replace the whole file (seemed simpler, but maybe not). But doing that, I find it better to avoid touching the code that the indico team/community wrote around the block that I'm adding. Eventually, I plan to make a PR on indico's side

jdkandersson:
> > Haven't finished this review yet, there are some broad questions about whether this should be a CLI or it should be in the charm source code. Let's address that first and then I'll finish the review
> 
> Sorry, it probably wasn't obvious from the PR description. I'm patching a file from the source code of indico ([here](https://github.com/indico/indico/blob/master/indico/cli/user.py) is the source). I decided to do it that way to avoid fiddling with their internal database structure.
> 
> So I want to do it in a way that is simple, ideally robust to some unrelated changes in indico and better, that breaks when there are changes that would affect this change. I thought of 3 ways to do it:
> 
> 1. wget the source and adding my bit at the end
> 2. wget the source and apply a git patch
> 3. replacing the whole file
> 
> I decided to replace the whole file (seemed simpler, but maybe not). But doing that, I find it better to avoid touching the code that the indico team/community wrote around the block that I'm adding. Eventually, I plan to make a PR on indico's side

So are we making any modifications from that source or is it just vending in that file? with no changes?

nrobinaubertin:
> So are we making any modifications from that source or is it just vending in that file? with no changes?

I'm adding the block with the command "autocreate" (https://github.com/canonical/indico-operator/blob/4bc932f871b6a494658c38fed25b489d84fbf9dc/files/src/indico/cli/user.py#L195)

jdkandersson:
> > So are we making any modifications from that source or is it just vending in that file? with no changes?
> 
> I'm adding the block with the command "autocreate"
> 
> https://github.com/canonical/indico-operator/blob/4bc932f871b6a494658c38fed25b489d84fbf9dc/files/src/indico/cli/user.py#L195

Would it be possible to move that to the charm source code?

gregory-schiano:
> > Haven't finished this review yet, there are some broad questions about whether this should be a CLI or it should be in the charm source code. Let's address that first and then I'll finish the review
> 
> Sorry, it probably wasn't obvious from the PR description. I'm patching a file from the source code of indico ([here](https://github.com/indico/indico/blob/master/indico/cli/user.py) is the source). I decided to do it that way to avoid fiddling with their internal database structure.
> 
> So I want to do it in a way that is simple, ideally robust to some unrelated changes in indico and better, that breaks when there are changes that would affect this change. I thought of 3 ways to do it:
> 
> 1. wget the source and adding my bit at the end
> 2. wget the source and apply a git patch
> 3. replacing the whole file
> 
> I decided to replace the whole file (seemed simpler, but maybe not). But doing that, I find it better to avoid touching the code that the indico team/community wrote around the block that I'm adding. Eventually, I plan to make a PR on indico's side

The risk here is that upgrading indico version could lead to incompatibility. Which means we'll have to pay attention to port the change on that file everytime we update indico.
A patch is a bit more flexible but still not ideal
Ideally we should submit this change to indico project to have it as part of the project itself.
Wouldn't it be possible to add an extra file independent that adds this new command ?

nrobinaubertin:
> Wouldn't it be possible to add an extra file independent that adds this new command ?

I thought of adding a new module in its own file. But I still have to load it therefore change the `core.py` file of indico's CLI

gregory-schiano:
> > Wouldn't it be possible to add an extra file independent that adds this new command ?
> 
> I thought of adding a new module in its own file. But I still have to load it therefore change the `core.py` file of indico's CLI

After investigating and discussing with Niels, Indico supports plugins that can extend CLI
I suggested to have a look at this as it will be less intrusive and more generic

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,49 @@
...
+
+    user_type = "user" if not grant_admin else "admin"
+    email = email.lower()
+    username = email
+    first_name = first_name or "unknown"
+    last_name = last_name or "unknown"
+    affiliation = affiliation or "unknown"
+
+    if User.query.filter(User.all_emails == email, ~User.is_deleted, ~User.is_pending).has_rows():
+        return
```
gregory-schiano, line 32: Shouldn't it be interesting to print something to give feedback ?
How is the return code managed ? If that's an error we should have >0

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if password == "":
+        return
```
gregory-schiano, line 35: Same as above

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if Identity.query.filter_by(provider="indico", identifier=username).has_rows():
+        click.secho("Username already exists", fg="red")
```
gregory-schiano, line 38: It doesn't return, so it'll still try to create the user ? Or it will update it ?
If it updates, better to be more specific in the message we print stating that the user will be updated

src/charm.py
```diff
@@ -689,6 +692,12 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
+        indico_env_config = self._get_indico_env_config(container)
+        for key in indico_env_config:
+            indico_env_config[key] = str(indico_env_config[key])
```
gregory-schiano, line 698: Perhaps using a dict comprehension ?
`return {env_var:str(env_value) for env_var, env_value in indico_env_config.items()}`

src/charm.py
```diff
@@ -905,6 +914,49 @@ def _has_secrets(self) -> bool:
         # check if the other end of a relation also supports secrets...
         return juju_version.has_secrets
 
+    def _add_user_action(self, event: ActionEvent) -> None:
+        """Add a new user to Indico.
+
+        Args:
+            event: Event triggering the add_user action.
```
gregory-schiano, line 921: Should probably be
`Event triggered by the add_user action`

src/charm.py
```diff
@@ -905,6 +914,49 @@ def _has_secrets(self) -> bool:
...
+            "user",
+            event.params["email"],
+            event.params["password"],
+        ]
+        if event.params.get("first-name"):
+            cmd.append(f"--first-name='{event.params.get('first-name')}'")
+        if event.params.get("last-name"):
+            cmd.append(f"--last-name='{event.params.get('last-name')}'")
+        if event.params.get("affiliation"):
+            cmd.append(f"--affiliation='{event.params.get('affiliation')}'")
```
gregory-schiano, line 938: I think some factoring is possible here using a small function to avoid some repetition ?

tests/unit/test_actions.py
```diff
@@ -0,0 +1,214 @@
...
+            "is-admin": True,
+        }
+
+        def event_store_failure(arg):
+            event.fail_message = arg
+
+        event.fail = event_store_failure
+
+        indico_env_config = charm._get_indico_env_config_str(container)
+        cmd = [
```
gregory-schiano, line 114: Perhaps add "expected_" prefix to the variable names ?

src/charm.py
nrobinaubertin: It's 6 lines.  
With a function definition (2 lines) + the calls (3 lines), we would gain 1 line. Is it worth it ?

plugins/autocreate/autocreate/cli.py
nrobinaubertin: I added an output message and exited with 1

plugins/autocreate/autocreate/cli.py
nrobinaubertin: I added an output message and exited with 1

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+    print(f"  First name: {user.first_name}")
+    print(f"  Family name: {user.last_name}")
+    print(f"  Email: {user.email}")
+    print(f"  Affiliation: {user.affiliation}")
+    if flags:
+        print(cformat("  Flags: {}".format(", ".join(flags))))
+    print()
+
+
+def _safe_lower(s):
```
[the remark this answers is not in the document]
nrobinaubertin: Outdated after I implemented it as a plugin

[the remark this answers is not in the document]
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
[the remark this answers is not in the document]
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
```diff
@@ -0,0 +1,515 @@
...
+from indico.modules.users import User
+from indico.modules.users.operations import create_user
+from indico.modules.users.util import search_users
+from indico.util.console import cformat, prompt_email, prompt_pass
+from indico.util.date_time import utc_to_server
+from terminaltables import AsciiTable
+
+
+@cli_group()
+def cli():
```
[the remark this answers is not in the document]
nrobinaubertin: Outdated after I implemented it as a plugin

files/src/indico/cli/user.py
nrobinaubertin: Outdated after I implemented it as a plugin

src/charm.py
jdkandersson: It is the standard way of building dictionaries in Python, just like list comprehensions

actions.yaml
jdkandersson: Do we have tests for each of the options? Each additional option raises the maintenance cost of the charm due to tests, keeping more code up to date etc

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,52 @@
...
+@cli.command("user")
+@click.argument("email", type=str)
+@click.argument("password", type=str)
+@click.option("--first-name", help="First name of the user", type=str)
+@click.option("--last-name", help="Last name of the user", type=str)
+@click.option("--affiliation", help="Affiliation of the user", type=str)
+@click.option("--admin/--no-admin", "grant_admin", is_flag=True, help="Grant admin rights")
+@click.pass_context
+def autocreate(ctx, email, password, first_name, last_name, affiliation, grant_admin):
+    """Create a new user non-interactively."""
```
jdkandersson, line 23: This doesn't look like it is covered by our linting, should we add this to be covered by our linting checks?

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,52 @@
...
+def cli():
+    """Create users non-interactively."""
+
+
+@cli.command("user")
+@click.argument("email", type=str)
+@click.argument("password", type=str)
+@click.option("--first-name", help="First name of the user", type=str)
+@click.option("--last-name", help="Last name of the user", type=str)
+@click.option("--affiliation", help="Affiliation of the user", type=str)
```
jdkandersson, line 19: Would it be worth coming up with a better description of this? The current description doesn't enhance my understanding of what `affiliation` means here

plugins/autocreate/autocreate/cli.py
```diff
@@ -0,0 +1,52 @@
...
+    """Create a new user non-interactively."""
+
+    email = email.lower()
+    username = email
+    first_name = first_name or "unknown"
+    last_name = last_name or "unknown"
+    affiliation = affiliation or "unknown"
+
+    if User.query.filter(User.all_emails == email, ~User.is_deleted, ~User.is_pending).has_rows():
+        click.secho('This user does already exist', fg='red')
```
jdkandersson, line 32: `This user already exists`?

plugins/autocreate/autocreate/plugin.py
```diff
@@ -0,0 +1,19 @@
+from indico.core import signals
+from indico.core.plugins import IndicoPlugin
+
+from autocreate.cli import cli
+from autocreate import _
```
jdkandersson, line 5: This is surprising, should we add a comment explaining why this is needed? I'm guessing there is some side-effect from importing autocreate?

plugins/autocreate/setup.cfg
```diff
@@ -0,0 +1,31 @@
+[metadata]
+name = indico-plugin-autocreate
+version = 3.2
+description = Extends the indico CLI to add a non-interactive way to create users
+long_description = file: README.md
+long_description_content_type = text/markdown; charset=UTF-8; variant=GFM
+url = https://github.com/canonical/indico-operator
+license = MIT
```
jdkandersson, line 8: Why is this licence MIT when the repo license is Apache 2.0? https://github.com/canonical/indico-operator/blob/main/LICENSE

plugins/autocreate/setup.cfg
```diff
@@ -0,0 +1,31 @@
...
+author = nrobinaubertin
```
jdkandersson, line 9: Should we make this a team author and email?

plugins/autocreate/setup.cfg
```diff
@@ -0,0 +1,31 @@
...
+author_email = niels.robin-aubertin@canonical.com
+classifiers =
+    Environment :: Plugins
+    Environment :: Web Environment
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
jdkandersson, line 16: Not 3.11?

plugins/autocreate/setup.cfg
```diff
@@ -0,0 +1,31 @@
...
+python_requires = >=3.9.0, <3.11
+install_requires =
+    indico>=3.2
+
+[options.entry_points]
+indico.plugins =
+    autocreate = autocreate.plugin:AutocreatePlugin
+
+[pydocstyle]
+ignore = D100,D101,D102,D103,D104,D105,D107,D203,D213
```
jdkandersson, line 31: What is the reason for these ignores please?

tests/unit/test_actions.py
```diff
@@ -0,0 +1,214 @@
...
+            {
+                "customization_sources_url": "https://example.com/custom",
+                "external_plugins": "git+https://example.git/#subdirectory=themes_cern",
+            }
+        )
+
+        charm: IndicoOperatorCharm = typing.cast(IndicoOperatorCharm, self.harness.charm)
+        charm._refresh_external_resources(MagicMock())
+
+        mock_exec.assert_any_call(
```
jdkandersson, line 53: Are these tests worth it since it looks like it is just repeating the implementation? Would it be better to cover this with integration tests?

jdkandersson (commented):
Any integration tests?

tests/unit/test_actions.py
nrobinaubertin: If I don't implement those, we will have to accept a lower coverage

plugins/autocreate/setup.cfg
nrobinaubertin: I could add it but it's not even the default on ubuntu 22.04.  
Are we running charms with it ?

plugins/autocreate/setup.cfg
nrobinaubertin: Removed.

plugins/autocreate/setup.cfg
nrobinaubertin: Sure, what is it ?

nrobinaubertin: Sure, which one should I use ?

jdkandersson: Maybe the `maintainers` from `metadata.yaml`? and `is-devops-team@canonical.com`?

plugins/autocreate/autocreate/cli.py
nrobinaubertin: I *think* that indico wants this field to be filled by a company name or a lab name.  
But it's just a guess on my side (I've copied indico's fields from their "user create" CLI)

plugins/autocreate/setup.cfg
jdkandersson: Up too you, I was just wondering whether there was some reason 3.11 was not included

tests/unit/test_actions.py
jdkandersson: It might be worth considering whether the tests add value (which they might, I haven't looked at them in detail). If it is just for the sake of coverage, we can add a `noqa` comment explaining why it isn't worth covering this part of the code

src/charm.py
gregory-schiano: And they're more efficient, I'd suggest to use them systematically (isn't it part of our guidelines ?)

src/charm.py
gregory-schiano: Ok agreed, let's keep it like this for now

plugins/autocreate/setup.cfg
gregory-schiano: Probably copy paste from other plugins, needs to be adapted

tests/unit/test_actions.py
gregory-schiano: If I'm not wrong this test wasn't added, but was already there and approved from a previous PR.
It just appears as a change because Niels had to split the test files due to pylint complaining of too many public methods.
I'd suggest to keep it since it's not the scope of this PR to review relevance of unit tests

indico.Dockerfile
```diff
@@ -28,6 +31,7 @@ RUN python3 -m pip install --no-cache-dir --no-warn-script-location --prefer-bin
     python-ldap \
     python3-saml \
     uwsgi
+RUN python3 -m pip install --no-cache-dir /srv/indico/plugins/autocreate
```
arturo-seijas, line 34: Please, put these in the `pip install`instruction above

nrobinaubertin: Done

plugins/autocreate/setup.cfg
nrobinaubertin: Done

plugins/autocreate/setup.cfg
nrobinaubertin: Fixed

plugins/autocreate/autocreate/plugin.py
nrobinaubertin: Removed

plugins/autocreate/autocreate/cli.py
nrobinaubertin: Fixed

plugins/autocreate/autocreate/cli.py
nrobinaubertin: The action changed to add-admin. Affiliation is gone

plugins/autocreate/autocreate/cli.py
nrobinaubertin: There are a lot of imports coming from indico or click. I'm not sure how to handle this if I add the plugin to our linting. Should I add those dependencies to the tox lint env ?

src/charm.py
nrobinaubertin: Replaced

actions.yaml
nrobinaubertin: I changed the action to an add-admin one

plugins/autocreate/autocreate/cli.py
jdkandersson: Can't we add this file to the linting checks in tox? Like change the `[vars]` to:

```tox
[vars]
src_path = {toxinidir}/src/
tst_path = {toxinidir}/tests/
cli_path = {toxinidir}/plugins/autocreate/
;lib_path = {toxinidir}/lib/charms/operator_name_with_underscores
all_path = {[vars]src_path} {[vars]tst_path} {[vars]cli_path}
```

jdkandersson:
Try to avoid `force-push` because it means that it is difficult to be able to review the changes since the last review. An alternative is using merge commits so that the history doesn't have to be revised

plugins/autocreate/autocreate/plugin.py
```diff
@@ -1,8 +1,8 @@
-from indico.core import signals
```
jdkandersson: Is this missing the copyright notice?

src/charm.py
```diff
@@ -694,9 +694,7 @@ def _get_indico_env_config(self, container: Container) -> Dict:
 
     def _get_indico_env_config_str(self, container: Container) -> Dict[str, str]:
         indico_env_config = self._get_indico_env_config(container)
-        for key in indico_env_config:
-            indico_env_config[key] = str(indico_env_config[key])
-        return indico_env_config
+        return {k: str(v) for k, v in indico_env_config.items()}
```
jdkandersson, line 697: Should we use more descriptive variable names than `k` and `v`?

tests/integration/test_charm.py
```diff
@@ -94,3 +95,25 @@ async def test_health_checks(app: Application):
             assert stdout.count("0/3") == 1
         else:
             assert stdout.count("0/3") == 2
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_add_admin(app: Application):
+    """Test the add-admin action.
```
jdkandersson, line 103: We have a standard for how to write the test docstrings, is it applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

tests/integration/test_charm.py
```diff
@@ -94,3 +95,25 @@ async def test_health_checks(app: Application):
...
+
+    Assume that the charm has already been built and is running.
+    """
+
+    assert app.units[0]  # type: ignore
```
jdkandersson, line 108: What is the reason for this ignore? We have a standard for documenting why any ignore statements were added: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

Also applies to the other `nosec`, `type: ignore` etc in this PR

tests/integration/test_charm.py
```diff
@@ -94,3 +95,25 @@ async def test_health_checks(app: Application):
...
+    email = "sample@email.com"
+    password = "somepassword"  # nosec
+
+    action: juju.action.Action = await app.units[0].run_action(  # type: ignore
+        "add-admin", email=email, password=password
+    )
+    await action.wait()
+    assert action.status == "completed"
+    assert action.results["user"] == email
+    assert f'Admin with email "{email}" correctly created' in action.results["output"]
```
jdkandersson, line 119: Would it be worth checking that the user is actually usable to do admin things?

plugins/autocreate/autocreate/plugin.py
jdkandersson: Also applicable to any other files without the copyright notice

plugins/autocreate/autocreate/cli.py
nrobinaubertin: Yes. As I said, if I do that, it's going to explode with import errors because it can't access indico's and click's libs. Should I add those dependencies to the tox lint env ?

jdkandersson: That probably makes sense to do, yes

src/charm.py
nrobinaubertin: Isn't standard for `key` and `value` ? Furthermore, in a tiny comprehension like this one, it's beneficial to have a small line (in length) to "see" it all at once.

plugins/autocreate/autocreate/plugin.py
nrobinaubertin: Added

tests/integration/test_charm.py
nrobinaubertin: I added a comment

tests/integration/test_charm.py
nrobinaubertin: I added a comment

tests/integration/test_charm.py
nrobinaubertin: I think that if we trust that indico's internals are doing the right thing (that is, creating an admin when I ask for it), we don't need to test that.

plugins/autocreate/autocreate/cli.py
nrobinaubertin: I added the plugin to the linting tests

plugins/autocreate/autocreate/__init__.py
```diff
@@ -1,3 +1 @@
-from indico.util.i18n import make_bound_gettext
-
-_ = make_bound_gettext("autocreate")
+"""Create users non-interactively."""
```
arturo-seijas, line 1: Can you add the license header too?

nrobinaubertin: Done

src/charm.py
```diff
@@ -905,6 +912,41 @@ def _has_secrets(self) -> bool:
...
+                cmd,
+                user="indico",
+                working_dir="/srv/indico",
+                environment=indico_env_config,
+            )
+            try:
+                output = process.wait_output()
+                event.set_results({"user": f"{event.params['email']}", "output": output})
+            except ExecError as ex:
+                logger.exception("Action add-admin failed")
```
merkata, line 943: You could also pass the exception to the logger, would that be helpful for debugging purposes if the action fails?

src/charm.py
gregory-schiano: Well I think it is worth it changing at least `k` to `env` as it's not a key, but an environment variable name.
`v` could stay as it's the `value` of the environment variable name.
The idea is to have a name which relates to its content

src/charm.py
nrobinaubertin: Done

arturo-seijas (commented):
Can you update the PR description to match the current action params?

tests/integration/test_charm.py
gregory-schiano: Well I'm a bit in between, since we created the plugin, there could be a bug in the plugin.
This remarks highlight more the lack of unittests for the plugin... but shoud we write unittests for the plugin as it's out the charm scope.
I'd say I'm fine with this as long as we don't extend the plugin more, if we extend the plugin it will be worth having the plugin on a separated repository, with its own tests and workflow.

src/charm.py
jdkandersson: There is cognitive overhead to reading short variable names because you have to figure out based on the context what they are. Adding 10 or so characters to a line of code to make it extremely obvious what is going on is a small price to pay for improved readability

`return {env_name: str(value) for env_name, value in indico_env_config.items()}`

tests/integration/test_charm.py
jdkandersson: The question is, how do we know whether we are using the Indico API correctly? Maybe a user is created, but there is some configuration we are missing or some mistake in how we interact with the API. You would likely manually try out this action to make sure the user being created is an admin user before we make this action available to our users. Why not automate that check?

jdkandersson: You can think about it a bit like this. The test currently checks just the the Indico API did something, it doesn't check that it did what we wanted. If you were a user of Indico, you would probably expect that we tried out this action to check that it actually creates a working admin user. If we're going to do that - just automate it in this test. That way we are guaranteeing that the feature our users depend on actually do what they need them to do.

jdkandersson: If you build a chair for someone, you probably sit on the chair to make sure it can actually take the load. You don't just visually inspect that it looks like a chair so it should be fine and you give it to someone, they sit on it and it breaks. Good way to have a poor chair building reputation

jdkandersson: Another perspective. Imagine you have to talk to our CEO in a few hours and they will ask you whether this feature works. You would probably do some thorough checks that it does, including making use of the admin user. Why wouldn't we apply this level of rigour to every feature we build and do so automatically rather than manually running commands and clicking around?

merkata: > If you build a chair for someone, you probably sit on the chair to make sure it can actually take the load. You don't just visually inspect that it looks like a chair so it should be fine and you give it to someone, they sit on it and it breaks. Good way to have a poor chair building reputation

With chairs it can get messy

image

gregory-schiano: Well I understand your arguments, but I'm still in between.
In our tests we don't ensure that ALL features of the application we're operating with this charm works. Usual test are just calling the homepage and expecting a 200.
This is very limited, and you effectively don't check that the connection to the DB works properly in some cases (in some application the homepage doesn't connect to the DB). Same goes to Redis for example where you perhaps don't trigger code path that calls redis...

You can't go too far, and must accept that you're building your operator on top of someone else's work whose responsibility is to ensure things works as designed.
Our tests must ensure that what we provide to the application as configuration and actions are as per contract (contract being what is in the documentation of the application). If the contract states that to create an admin you must call "add-admin" on the CLI, our tests must ensure that we effectively called the "add-admin" on the CLI. Then we trust the application to properly create an Admin user when this CLI is called.
Unfortunately here we're the one who wrote the contract, and we didn't cover that contract with proper testing.

I must admit that also going further than just trusting the provider that the documentation is ok and the implementation is ok could be a mistake. But we're not talking here about critical business (health, defense, rockets) we must accept a certain level of trust (we don't test that github is working properly, that the internet provider of github is working properly, that the docker images of github are the one we expect, that all the packages installed in our OCI image contains what we expect, that the machine github uses is effectively the one we expect, that the CPU build is the design expected, ...)

To reply to you questions:
`The question is, how do we know whether we are using the Indico API correctly?`
Well once more the contract is "call add-admin if you want to have an admin user", trust me bro, I'm taking care of my part of the contract.
But as previously said, we're the provider and the user of that contract, and we know the contract is not tests. Hence my first comment about... this should be on a different repo, covered with its own unittests.
`Why not automate that check?`
We must make sure that the test covers the contract, if the contract tell you need to have pass hostname, username and password of the BDD to ensure the application accesses the DB, and you must call the CLI with username, email and password. Our test must ensure this contract is followed properly, and trust the provider of the contract that it will achieve what's described in the documentation.
`If you build a chair for someone, you probably sit on the chair to make sure it can actually take the load.`
Surely, but you won't analyze the wood with microscope to ensure that the provider effectively delivered you "Oak wood" on which you base your load design.
`Why wouldn't we apply this level of rigour to every feature we build and do so automatically rather than manually running commands and clicking around?`
Taking your question... what would you do if the feature described in the documentation of the product doesn't do what it's supposed to do ?
If that feature is broken, well you won't be able to show this feature to your CEO because it's unlikely you'll fix the application yourself. You'll open a ticket to the application provider and tell to your CEO... we'll this application has a bug, I can't show the add-admin feature.
I admit it's better to know that this feature doesn't work upfront, but will you then cover with integration test ALL the feature of the application you'll show to your CEO ?

gregory-schiano: And don't get me wrong, there are situations where it's better to test the expected outcome on the application. Because you know the contract is not always 100% clear, and you could be wrong on interpreting it.
If I take as an example SAML, there are many ways you could pass the wrong parameter from your operator and if you just check that "Variable X" is there, that won't be enough as the format of "X" could be wrong.
In an ideal world, the contract wouldn't leave any ambiguity and you could just rely on it, but we're not in this ideal world

jdkandersson: The question is more the degree to which you test the contract. Yes you don't have to put the chair's wood under a microscope for each chair, you would subject the chair to an expected load before you ship it to your customer. I used to work in manufacturing food and the excuse of "the contract with the supplier says this" is not going to satisfy the regulator or the customers you just made sick if you ship a product that doesn't meet standards because of a mistake by the supplier. You are accountable to your customer for what you ship, nobody else. Contracts you have with suppliers don't matter to the customer or regulator you sold the product to.

Whilst we are not directly in the critical business like defense or rocket science, I would bet that those industries use our products and we would expect our products to be of a standard that they could be used by those critical industries. The things we build are building blocks or used for the building blocks for many other industries, just like Python is. If Canonical makes a mistake building Ubuntu, the impact would be widespread. That mistake could originate in some of the tools used during the build process as much as it could originate from the Ubuntu source code itself.

What is reasonable in this case is to make a CLI call or interact with Indico in some simple way using the created credentials just to make sure that it worked. It could just be using that credential to query some information that only the admin can. An example where something like this would have helped: the `upload-charm-docs` migration feature was initially merged without these kind of end-to-end tests. We thought we were using the contract with our dependencies correctly, but it turned out that the contract changes when it is run the way our users would run the action and the feature broke. If we had an end-to-end the test ran the action and checked the outcome of the action, we would have caught this bug before having an impact to our team and potentially to users.

There is a question of which of the Indico features we should check. I agree that we can't go too far, I don't think that we go far enough yet. We probably don't need to test features that are just inherent to the application, like that an event can be scheduled or a specific website works. When we explicitly provide an action on the charm, we should be sure that it works because we are making that feature available to our users and are accountable for it. If we are providing an integration with the database and redis, it would be concerning if we don't check that the integration works. It isn't about checking that Indico and redis have been coded properly, or that the network works, or that GitHub works. It is about checking that the feature we guarantee for our users, that you can integrate Indico with a database or redis, works. How would we know without doing at least one things that requires it?

And yes we can move it to a different repo, there should be checks there that make sure it works as expected. Once we have provided a guarantee there, we can be more relaxed about using it because that is within what our team guarantees.

On the question of whether you cover what you show to the CEO, whatever features you claim to deliver in a product should be covered by tests. The Indico charm doesn't claim to deliver an event experience, it claims to make it easier to operate an existing service like Indico through providing actions like creating an admin user. If we claim to deliver this feature, we should test it. We don't white label Indico so we don't need to test things in Indico we don't directly claim to deliver to our customers.

jdkandersson: I should add this, it could be that adding an integration test to exercise the admin user is complex or expensive to write in which case it could be that it isn't worth doing in certain cases

jdkandersson: I propose we move this discussion to this PR adding a standard to our contributing guide: https://github.com/canonical/indico-operator/pull/220#discussion_r1091363449

src/charm.py
nrobinaubertin: changed

amandahla:
> > > Wouldn't it be possible to add an extra file independent that adds this new command ?
> > 
> > 
> > I thought of adding a new module in its own file. But I still have to load it therefore change the `core.py` file of indico's CLI
> 
> After investigating and discussing with Niels, Indico supports plugins that can extend CLI I suggested to have a look at this as it will be less intrusive and more generic

Just to add another option to consider, it could be possible to pass input to the "indico user create --admin" command using `expect`:
https://phoenixnap.com/kb/linux-expect

So wouldn't be necessary to change anything in Indico.

tests/integration/test_charm.py
nrobinaubertin: After discussions, it was decided that further tests would not be worth it (as part of this PR). 
A new issue https://github.com/canonical/indico-operator/issues/230 has been created as follow-up.

nrobinaubertin:
> Just to add another option to consider, it could be possible to pass input to the "indico user create --admin" command using expect

Yes, this was considered during the discussions that lead to a conclusion to [this thread](https://github.com/canonical/indico-operator/pull/220#discussion_r1091363449).  
The answer of David was that even using that way of doing things would call for more e2e testing.

plugins/autocreate/autocreate/cli.py
```diff
@@ -45,8 +51,10 @@ def create_admin(ctx, email, password):
     )
     user.is_admin = True
 
-    db.session.add(user)
-    db.session.commit()
+    # db.session has add()
+    db.session.add(user)  # pylint: disable=no-member
```
jdkandersson, line 55: What is the reason for mypy complaining here? Does it think that it could be `None`?

plugins/autocreate/autocreate/cli.py
```diff
@@ -63,4 +71,10 @@ def create_admin(ctx, email, password):
         click.secho("Admin was not correctly created", fg="red")
         ctx.exit(1)
 
-    click.secho(f'Admin with email "{res.pop().email}" correctly created', fg="green")
+    user = res.pop()
+
+    if not user.is_admin:
```
jdkandersson, line 76: Nice bit of safety

plugins/autocreate/autocreate/plugin.py
```diff
@@ -15,5 +19,7 @@ def init(self):
         super().init()
         self.connect(signals.plugin.cli, self._extend_indico_cli)
 
-    def _extend_indico_cli(self, sender, **kwargs):
+    # We need those arguments when extending indico's CLI
+    # pylint: disable=unused-argument
+    def _extend_indico_cli(self, *args, **kwargs):
```
jdkandersson, line 24: What about `*_` and `**_`, does that resolve the pylint warning?

plugins/autocreate/autocreate/cli.py
nrobinaubertin:
```
plugins/autocreate/autocreate/cli.py:55:4: E1101: Instance of 'scoped_session' has no 'add' member (no-member)
plugins/autocreate/autocreate/cli.py:57:4: E1101: Instance of 'scoped_session' has no 'commit' member (no-member)
```

plugins/autocreate/autocreate/plugin.py
nrobinaubertin: changed

plugins/autocreate/autocreate/cli.py
jdkandersson: Ah yes, I'm familiar with those, they are unfortunate that pylint gets it wrong

jdkandersson (dismissed):
Looking good, still seems to be some issues with the lint command

Thanks for your resilience with all the feedback!

nrobinaubertin:
> Looking good, still seems to be some issues with the lint command

The issue here is that the lint tests are running on focal (python 3.8) and the pypi package for indico 3.2 is not compatible with it.
I think that the best course of action is to remove the plugin from the lint tests.

.github/workflows/test.yaml
```diff
@@ -7,3 +7,13 @@ jobs:
   unit-tests:
     uses: canonical/operator-workflows/.github/workflows/test.yaml@main
     secrets: inherit
+  plugins-test:
```
arturo-seijas, line 10: Is it possible to do this just by changing the tox lint target?

nrobinaubertin: What do you mean by that ?

arturo-seijas (approved):
LGTM

amandahla (approved):
LGTM

jdkandersson:
The integration tests failed...

jdkandersson:
If the lint checks fail for Ubuntu 20.04, does the `add-admin` action actually work on Python 3.8? It looks like it wasn't possible to install indico on Python 3.8?

PR #225, opened 2023-01-30
Fix dependencies needed for package installation

arturo-seijas:
Fix dependencies needed for python package installation and pin Indico version to 3.2.0, last version does not require Postgres 13.
As part of this, the Indico image is built as a single stage build. Multistage doesn't provide any advantage.

### DIFF
--- a/indico.Dockerfile
+++ b/indico.Dockerfile
@@ -22,7 +22,7 @@ ENV UWSGI_EMBED_PLUGINS=stats_pusher_statsd
 
 USER indico
 RUN python3 -m pip install --no-cache-dir --no-warn-script-location --prefer-binary \
-    indico==3.2 \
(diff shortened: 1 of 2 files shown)


nrobinaubertin (dismissed):

indico.Dockerfile
```diff
@@ -3,52 +3,23 @@ FROM ubuntu:jammy as builder
 ARG indico_gid=2000
```
mthaddon, line 3: Seems like the line above needs updating now we're no longer doing multi-stage builds?

```FROM ubuntu:jammy as builder```

mthaddon (approved):
LGTM, thx

gtrkiller (approved):
LGTM

merkata (approved):

PR #226, opened 2023-01-31
Fix SSO and missing packages issues

arturo-seijas:
* Fix SSO integration with Canonical's SAML server by renaming the settings dict key
* Install packages as root to guarantee they are executable by any user


merkata (approved):
Short, nice and clear. Thanks!

gtrkiller (approved):
LGTM

PR #236, opened 2023-02-09
Adjust memberOf casing to fix group membership

kewisch:
This should fix group membership checks, our ldap server uses the default `memberOf` instead of `memberof`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -659,7 +659,7 @@ class IndicoOperatorCharm(CharmBase):
                     "gid": "cn",
                     "group_base": "dc=canonical,dc=com",
                     "group_filter": "(objectClass=groupofnames)",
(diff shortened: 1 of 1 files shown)


jdkandersson (approved):

jdkandersson:
Looks like tests are failing

kewisch:
Do you have an idea what it might be? I've skimmed over them and it doesn't look related. Possibly it is not using the secrets because I am not a repo owner?

arturo-seijas (approved):
LGTM

jdkandersson:
> Do you have an idea what it might be? I've skimmed over them and it doesn't look related. Possibly it is not using the secrets because I am not a repo owner?

Yes it looks like this is a secrets issue. I'll merge this into a separate branch and then we can run the tests again

PR #244, opened 2023-02-22
Isd 392 indico fix the issues found by trivy scanning

merkata:
This PR addresses the issues found during a trivy scan. Python dependencies are updated upstream on Indico in a patch release and I've bumped the Indico version, I've ignored a disputed vulnerability and the pebble CVEs until they get addressed in https://github.com/canonical/pebble/pull/194.


indico_rock/rockcraft.yaml
```diff
@@ -53,7 +53,7 @@ parts:
     override-build: |
       craftctl default
       UWSGI_EMBED_PLUGINS=stats_pusher_statsd python3 -m pip install --no-cache-dir --no-warn-script-location --prefer-binary \
-        indico==3.2.0 \
+        indico==3.2.2 \
```
arturo-seijas, line 56: Iirc, this is not compatible with postgresql 12, the last version available in our machine charms

arturo-seijas (changes requested):
Indico 3.2.2 doesn't seem to guarantee compatibility with postgresql 12

indico_rock/rockcraft.yaml
merkata: Just to make sure I understand it - if we have Indico deployed at v3.2.2 in k8s and the backend is postgresql 12 running on machine, that would fail? If Indico follows the semver convention that should be only a patch without any breaking changes.

arturo-seijas: Exactly. They've added validation for the db version. Maybe it can be disabled, I'm not sure, but I don't know if that will have any consequences. From the [release notes, version 3.2.1](https://docs.getindico.io/en/latest/changelog/#id8):
```
Require at least Postgres 13 during new installations. This check can be forced on older Postgres versions (11+ should work), but we make no guarantees that nothing is broken (the latest version we test with is 12) ([#5503](https://github.com/indico/indico/pull/5503))
```

merkata: Thanks for digging that out, it is clear now :)

merkata: Alright, looking at the pipeline, `3.2.2` resolves all issues - good. We cannot use it still though. I'll revert and add the findings to `.trivyignore` with a comment.

arturo-seijas (approved):
LGTM

amandahla (approved):
LGTM

yanksyoon (approved):
LGTM :)
Out of curiosity, do we have processes to alert us when dependencies become resolved or would it have to be manual polling for now?
I assume this issue is a blocker and hence this PR?

merkata:
> LGTM :) Out of curiosity, do we have processes to alert us when dependencies become resolved or would it have to be manual polling for now? I assume this issue is a blocker and hence this PR?

That's a good point, we don't have a process for alerting, we can think further about this. This issue is just responding to the implementation of trivy scan and the findings - at this point we are ignoring some, which is OK, as we are all aware which these findings are and why we are ignoring them. It is not a blocker. The PR still verified that bumping to Indico 3.2.2 would fix the reported CVEs, it is just that we cannot yet bump Indico to that version (due to the requirement of postgresql >= 13).

Issue #242, opened 2023-02-20
Increase timeout for email and import endpoints


kewisch:
### Bug Description

I mentioned this on MM, but filing here to track it somewhere. The following URLs have a 504 gateway timeout, though they are long running operations. This causes some undefined errors in Indico which causes emails to be sent many times, or a potentially incomplete import operation.

POST https://events.canonical.com/event/NNN/manage/registration/NNN/registrations/email 

POST https://events.canonical.com/event/NNN/manage/registration/NNN/registrations/import

If you could bump the timeout for this to several minutes this would be helpful.

### To Reproduce

One simple way to test this is to import >500 users into an event registration using CSV.

### Environment

events.canonical.com
events.staging.canonical.com

### Relevant log output

```shell
--
```

### Additional context

Once changed, we should confirm this works as expected, I'm not sure if there is also a client side read timeout involved in addition. The message I get is definitely server side though.

amandahla:
The timeout was changed for:
```
POST https://events.staging.canonical.com/event/NNN/manage/registration/NNN/registrations/email 
POST https://events.staging.canonical.com/event/NNN/manage/registration/NNN/registrations/import
```
Test: 800 participants were successfully imported.

PR #245, opened 2023-02-24
fix(ISD-405): Increase proxy read timeout in NGINX

amandahla:
There are some long operations (import registrants, for example) that require more time to finish. 

This PR changes the proxy_read_timeout (default 60s) to 300s (5 minutes) to prevent the user to receive an HTTP 504 Gateway Timeout error.

Local tests showed that this was enough to import ~10000 registrants at once (took ~3min).

### DIFF
--- a/nginx_rock/etc/nginx.conf
+++ b/nginx_rock/etc/nginx.conf
@@ -58,6 +58,7 @@ http {
       proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
       proxy_set_header X-Forwarded-Host $http_host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_read_timeout 300;
       proxy_pass http://localhost:8081;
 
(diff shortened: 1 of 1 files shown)


gtrkiller (dismissed):
LGTM

mthaddon:
Is it possible to apply this change only to specific URLs? Increasing the timeout globally risks tying up workers for longer, so it would be good to restrict to the URLs we expect to need it if possible.

arturo-seijas (dismissed):
LGTM

amandahla:
> Is it possible to apply this change only to specific URLs? Increasing the timeout globally risks tying up workers for longer, so it would be good to restrict to the URLs we expect to need it if possible.

Good point, I'm considering only the two URLs mentioned in the story (it might be more) so created a location for adding the custom proxy_read_timeout there.

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

amandahla:
Just relating: one of the steps to fix #242

PR #246, opened 2023-03-02
ISD-382: Fix the issues found by OWASP scanning

amandahla:
* Fix Content Security Policy (CSP) Header Not Set alert
* Add ZAP.md with reasons why some alerts are ignored
* Add common_headers.conf file to NGINX config


ZAP.md
```diff
@@ -0,0 +1,34 @@
...
+The Ajax Spider scan is not a good fit for the integration test.
+
+## 90033    IGNORE  (Loosely Scoped Cookie)
+
+Specifying Domain is less restrictive than omitting it.
+Read more in [Cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
+
+## 10049-1 IGNORE  (Non-Storable Content)
+
+The response does contain sensitive, personal or user-specific information.
```
jdkandersson, line 30: Is this intended to say that the response does include the specific information? Is this potentially a concern?

zap_rules.tsv
```diff
@@ -0,0 +1,12 @@
...
+# Only the rule identifiers are used - the names are just for info
+# You can add your own messages to each rule by appending them after a tab on each line.
+10202   IGNORE  (Absence of Anti-CSRF Tokens)
+10031   IGNORE  (User Controllable HTML Element Attribute (Potential XSS))
+10049-3 IGNORE  (Storable and Cacheable Content)
+90003   IGNORE  (Sub Resource Integrity Attribute Missing)
+10109   IGNORE  (Modern Web Application)
+90033   IGNORE  (Loosely Scoped Cookie)
+10049-1 IGNORE  (Non-Storable Content)
+10027   IGNORE  (Information Disclosure - Suspicious Comments)
```
jdkandersson, line 12: Final new line?

zap_rules.tsv
```diff
@@ -0,0 +1,12 @@
+# zap-baseline rule configuration file
+# Change WARN to IGNORE to ignore rule or FAIL to fail if rule matches
+# Only the rule identifiers are used - the names are just for info
+# You can add your own messages to each rule by appending them after a tab on each line.
+10202   IGNORE  (Absence of Anti-CSRF Tokens)
```
jdkandersson, line 5: This seems to duplicate ZAP.md? Is it possible to add comments in this file with the reason for disabling the rules?

amandahla: This file is used by ZAP action so we can set each rule that should be ignored (or any other action) and I believe is not possible to add comments.

ZAP.md
amandahla: This alert https://www.zaproxy.org/docs/alerts/10049-1/ is about not caching content. In that case, the reason to not cache is because contains sensitive information (login, events or registrants data, etc.).

zap_rules.tsv
jdkandersson: It looks like there is a comment right above this line?

ZAP.md
```diff
@@ -14,7 +14,7 @@ The response does not contain sensitive, personal or user-specific information.
 
 ## 10049-3 IGNORE  (Sub Resource Integrity Attribute Missing)
 
-To be worked.
+SRI is not needed in this case.
```
jdkandersson, line 17: Would it be worth noting why?

ZAP.md
jdkandersson: I'm still confused, is this supposed to indicate why this rule is being disabled? If so, I don't follow that the sentence (including the updated version) answers that question?

zap_rules.tsv
amandahla: I had to remove it, it was not working :(

ZAP.md
amandahla: Actually, by investigating this [code](https://github.com/zaproxy/zap-extensions/blob/main/addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/CacheableScanRule.java#L130) responsible for how ZAP raises this informational alert, I realized I misunderstood.

This alert is being raised for paths /admin and /rooms because Indico redirects to /admin/settings and /rooms/book.

I'll update the description explaining what is happening and there is no action regarding cache to do about it since the redirect is expected.

ZAP.md
amandahla: Sure, I'll use Weii's explanation.

jdkandersson (approved):
Looks fine, the CI still seems to be failing

arturo-seijas (approved):
LGTM

merkata (approved):
Nice

PR #248, opened 2023-03-08
feat(ISD-260): Add anonymize-user action

amandahla:
Add anonymize-user action by using the same idea of adding an admin user: extends the CLI to anonymize a user using the e-mail list separated by a comma as a parameter.

Fields affected:
User data:
- ['affiliation', 'email', 'secondary_emails', ] = ""
- ['favorite_users', 'favorite_categories', 'identities'] = empty
- ['old_api_keys',] = empty
- ['first_name', 'last_name', 'phone', 'address',] = hash

Event Registration data:
- All text/textarea fields = hash
- Email = hash + "@invalid.invalid"
- Phone = (+00) 0000000
- First name = hash
- Last name = hash
- Email = hash

Reference:
https://github.com/bpedersen2/indico-cron-advanced-cleaner/

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -16,3 +16,4 @@ anonymize-user:
     email:
       type: string
       description: User email (or a list of emails separated by comma). Maximum of 50 emails.
+  required: [email]
--- a/actions.yaml
+++ b/actions.yaml
@@ -15,4 +15,4 @@ anonymize-user:
   params:
     email:
       type: string
-      description: User email.
+      description: User email (or a list of emails separated by comma)
--- a/actions.yaml
+++ b/actions.yaml
@@ -11,7 +11,7 @@ add-admin:
       description: User password.
   required: [email, password]
 anonymize-user:
-  description: Anonymize user in Indico
+  description: Anonymize stored personal data to facilitate GDPR compliance
   params:
     email:
       type: string
--- a/actions.yaml
+++ b/actions.yaml
@@ -15,4 +15,4 @@ anonymize-user:
   params:
     email:
       type: string
-      description: User email (or a list of emails separated by comma)
+      description: User email (or a list of emails separated by comma). Maximum of 50 emails.
--- a/actions.yaml
+++ b/actions.yaml
@@ -10,3 +10,9 @@ add-admin:
       type: string
       description: User password.
   required: [email, password]
+anonymize-user:
+  description: Anonymize user in Indico
+  params:
+    email:
+      type: string
+      description: User email.
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Create users non-interactively."""
+
+import click
+from hashlib import sha512
+from indico.cli.core import cli_group
+from indico.core.db import db
+from indico.modules.auth import Identity
+from indico.modules.users import User
+from indico.modules.users.operations import create_user
+from indico.modules.users.util import search_users
+from indico.util.date_time import now_utc
+
+
+@cli_group(name="anonymize")
+def cli():
+    """Anonymize users non-interactively."""
+
+# Extracted from:
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def anonymize_deleted_user(user):
+    """Anonymize user by erasing specific attributes."""
+
+    _anon_attrs = ['first_name', 'last_name', 'phone', 'address',]
+    _clear_attrs_str = ['affiliation', 'email', 'secondary_emails', ]
+    _clear_attrs_set = ['favorite_users', 'favorite_categories', 'identities']
+    _clear_attrs_list = ['old_api_keys',]
+
+
+    for attr in _clear_attrs_str:
+        setattr(user, attr, '')
+
+    for attr in _clear_attrs_set:
+        setattr(user, attr, set())
+
+    for attr in _clear_attrs_list:
+        setattr(user, attr, list())
+
+    for attr in _anon_attrs:
+        val = getattr(user, attr)
+        new_hash = sha512(val.encode('utf-8')+now_utc().isoformat().encode('utf-8')).hexdigest()[:12]
+        setattr(user, attr, new_hash)
+
+@cli.command("user")
+@click.argument("email", type=str)
+@click.pass_context
+def anonymize_user(ctx, email):
+    """Anonymize user non-interactively."""
+
+    email = email.lower()
+
+    if email == "":
+        click.secho("E-mail should not be empty", fg="red")
+        ctx.exit(1)
+
+    users = User.query.filter(User.all_emails == email)
+    if not users.has_rows():
+        click.secho("User not found", fg="red")
+        ctx.exit(1)
+    user = user.first()
+    # First we mark as deleted so won't appear in any form
+    user.is_deleted = True
+    anonymize_deleted_user(user)
+    db.session.commit()  # pylint: disable=no-member
+
+    # Validate the changes
+    res = search_users(
+        exact=True,
+        include_deleted=True,
+        include_pending=False,
+        include_blocked=False,
+        external=False,
+        allow_system_user=False,
+        email=email,
+    )
+
+    user = res.pop()
+
+    if not user.is_deleted:
+        click.secho("User was not anonymized", fg="red")
+        ctx.exit(1)
+
+    click.secho(f'User with email "{email}" correctly anonymized', fg="green")
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -15,10 +15,24 @@ from indico.modules.users.util import search_users
 from indico.util.date_time import now_utc
 import re
 
-CLEAN_ATTRS_STR = ['affiliation', 'email', 'secondary_emails', ]
-CLEAN_ATTRS_SET= ['favorite_users', 'favorite_categories', 'identities']
-CLEAN_ATTRS_LIST = ['old_api_keys',]
-CLEAN_ATTRS_ANON = ['first_name', 'last_name', 'phone', 'address',]
+
+EMPTY_LIST = lambda _: list()
+EMPTY_SET = lambda _: set()
+EMPTY_VAL = lambda _: ''
+HASH_VAL = lambda val: _hash(val)
+CLEAN_ATTRS = {
+    'affiliation' : EMPTY_VAL,
+    'email': EMPTY_VAL,
+    'secondary_emails': EMPTY_VAL,
+    'favorite_users': EMPTY_SET,
+    'favorite_categories': EMPTY_SET,
+    'identities': EMPTY_SET,
+    'old_api_keys': EMPTY_LIST,
+    'first_name': HASH_VAL,
+    'last_name': HASH_VAL,
+    'phone': HASH_VAL,
+    'address': HASH_VAL,
+}
 
 @cli_group(name="anonymize")
 def cli():
@@ -31,10 +45,12 @@ def _hash(val: str) -> str:
 
     Args:
         val: The string from the hash should be created
+    Returns:
+        First 12 characters from encoded data in hexadecimal format
     """
     return sha512(val.encode('utf-8')+now_utc().isoformat().encode('utf-8')).hexdigest()[:12]
 
-# Extracted from:
+# Based on:
 # https://github.com/bpedersen2/indico-cron-advanced-cleaner/
 def anonymize_deleted_user(user: User):
     """Anonymize user by erasing specific attributes.
@@ -42,30 +58,17 @@ def anonymize_deleted_user(user: User):
     Args:
         user: Indico user
     """
+    for attr, anonymize_val in CLEAN_ATTRS.items():
+        current_val = getattr(user, attr)
+        setattr(user, attr, anonymize_val(current_val))
 
-    for attr in CLEAN_ATTRS_STR:
-        setattr(user, attr, '')
-
-    for attr in CLEAN_ATTRS_SET:
-        setattr(user, attr, set())
-
-    for attr in CLEAN_ATTRS_LIST:
-        setattr(user, attr, list())
-
-    for attr in CLEAN_ATTRS_ANON:
-        setattr(user, attr, _hash(getattr(user, attr)))
-
-# Extracted from:
-# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
-def anonymize_registrations(user: User):
-    """Anonymize user by erasing registrations attributes.
+def anonymize_registration(registration: Registration):
+    """Anonymize registration by changing specific attributes.
 
     Args:
-        user: Indico user
+        registration (Registration): User registration_
     """
-    registrations = Registration.query.filter(Registration.user_id == user.id)
-    for registration in registrations:
-        for fid, rdata in registration.data_by_field.items():
+    for fid, rdata in registration.data_by_field.items():
             fieldtype = RegistrationFormField.get(oid=fid).input_type
             if fieldtype in ('text', 'textarea'):
                 rdata.data = _hash(rdata.data)
@@ -79,13 +82,26 @@ def anonymize_registrations(user: User):
             elif fieldtype == 'country':
                 # Keep country for now
                 pass
-        # other field types are not touched (choice, mulitchoice, radio)
+        # other field types are not touched (choice, multiple choice, radio)
 
-        registration.first_name = _hash(registration.first_name)
-        registration.last_name = _hash(registration.last_name)
-        registration.email = _hash(registration.email)
-        if registration.user:
-            registration.user = None
+    registration.first_name = _hash(registration.first_name)
+    registration.last_name = _hash(registration.last_name)
+    registration.email = _hash(registration.email)
+    if registration.user:
+        registration.user = None
+
+# Based on:
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def anonymize_registrations(user: User):
+    """Anonymize user by erasing registrations attributes.
+
+    Args:
+        user: Indico user
+    """
+    registrations = Registration.query.filter(Registration.user_id == user.id)
+    for registration in registrations:
+        anonymize_registration(registration)
+    # db.session has commit()
     db.session.commit()  # pylint: disable=no-member
 
 @cli.command("user")
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -4,13 +4,14 @@
 
 """Anonymize users non-interactively."""
 
+import uuid
+
 import click
 from indico.cli.core import cli_group
 from indico.core.db import db
 from indico.modules.events.registration.models.form_fields import RegistrationFormField
 from indico.modules.events.registration.models.registrations import Registration
 from indico.modules.users import User
-import uuid
 
 CLEAN_ATTRS = {
     "affiliation": str,
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -71,7 +71,7 @@ def anonymize_registration(registration: Registration):
         if fieldtype in ("text", "textarea"):
             rdata.data = _hash(rdata.data)
         elif fieldtype == "email":
-            rdata.data = _hash(rdata.data) + "@invalid.invalid"
+            rdata.data = f"{_hash(rdata.data)}@invalid.invalid"
         elif fieldtype == "phone":
             rdata.data = "(+00) 0000000"
         elif fieldtype == "date":
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -26,7 +26,7 @@ def cli():
 
 # Extracted from:
 # https://github.com/bpedersen2/indico-cron-advanced-cleaner/
-def _hash(val: str):
+def _hash(val: str) -> str:
     """Create hash of val.
 
     Args:
@@ -107,8 +107,8 @@ def anonymize_user(ctx, email):
 
     users = User.query.filter(User.all_emails == email)
     if not users.has_rows():
-        click.secho("User not found", fg="red")
-        ctx.exit(1)
+        click.secho("User not found", fg="yellow")
+        ctx.exit(0)
     user = users.first()
 
     # We mark as deleted so won't appear in search users forms
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -4,39 +4,32 @@
 
 """Anonymize users non-interactively."""
 
-import click
 from hashlib import sha512
+
+import click
 from indico.cli.core import cli_group
 from indico.core.db import db
-from indico.modules.events.registration.models.registrations import Registration
 from indico.modules.events.registration.models.form_fields import RegistrationFormField
+from indico.modules.events.registration.models.registrations import Registration
 from indico.modules.users import User
-from indico.modules.users.util import search_users
 from indico.util.date_time import now_utc
-import re
 
-
-EMPTY_LIST = lambda _: list()
-EMPTY_SET = lambda _: set()
-EMPTY_VAL = lambda _: str
-HASH_VAL = lambda val: _hash(val)
 CLEAN_ATTRS = {
-    'affiliation' : EMPTY_VAL,
-    'email': EMPTY_VAL,
-    'secondary_emails': EMPTY_VAL,
-    'favorite_users': EMPTY_SET,
-    'favorite_categories': EMPTY_SET,
-    'identities': EMPTY_SET,
-    'old_api_keys': EMPTY_LIST,
-    'first_name': HASH_VAL,
-    'last_name': HASH_VAL,
-    'phone': HASH_VAL,
-    'address': HASH_VAL,
+    "affiliation": str(),
+    "email": str(),
+    "secondary_emails": str(),
+    "favorite_users": set(),
+    "favorite_categories": set(),
+    "identities": set(),
+    "old_api_keys": [],
 }
+HASH_ATTRS = ["first_name", "last_name", "phone", "address"]
+
 
 @cli_group(name="anonymize")
 def cli():
     """Anonymize users non-interactively."""
+
 
 # Extracted from:
 # https://github.com/bpedersen2/indico-cron-advanced-cleaner/
@@ -48,7 +41,8 @@ def _hash(val: str) -> str:
     Returns:
         First 12 characters from encoded data in hexadecimal format
     """
-    return sha512(val.encode('utf-8')+now_utc().isoformat().encode('utf-8')).hexdigest()[:12]
+    return sha512(val.encode("utf-8") + now_utc().isoformat().encode("utf-8")).hexdigest()[:12]
+
 
 # Based on:
 # https://github.com/bpedersen2/indico-cron-advanced-cleaner/
@@ -60,7 +54,11 @@ def anonymize_deleted_user(user: User):
     """
     for attr, anonymize_val in CLEAN_ATTRS.items():
         current_val = getattr(user, attr)
-        setattr(user, attr, anonymize_val(current_val))
+        setattr(user, attr, anonymize_val)
+    for attr in HASH_ATTRS:
+        current_val = getattr(user, attr)
+        setattr(user, attr, _hash(current_val))
+
 
 def anonymize_registration(registration: Registration):
     """Anonymize registration by changing specific attributes.
@@ -70,16 +68,16 @@ def anonymize_registration(registration: Registration):
     """
     for fid, rdata in registration.data_by_field.items():
         fieldtype = RegistrationFormField.get(oid=fid).input_type
-        if fieldtype in ('text', 'textarea'):
+        if fieldtype in ("text", "textarea"):
             rdata.data = _hash(rdata.data)
-        elif fieldtype == 'email':
-            rdata.data = _hash(rdata.data) + '@invalid.invalid'
-        elif fieldtype == 'phone':
-            rdata.data = '(+00) 0000000'
-        elif fieldtype == 'date':
+        elif fieldtype == "email":
+            rdata.data = _hash(rdata.data) + "@invalid.invalid"
+        elif fieldtype == "phone":
+            rdata.data = "(+00) 0000000"
+        elif fieldtype == "date":
             # Keep dates for now
             pass
-        elif fieldtype == 'country':
+        elif fieldtype == "country":
             # Keep country for now
             pass
         # other field types are not touched (choice, multiple choice, radio)
@@ -89,6 +87,7 @@ def anonymize_registration(registration: Registration):
     registration.email = _hash(registration.email)
     if registration.user:
         registration.user = None
+
 
 # Based on:
 # https://github.com/bpedersen2/indico-cron-advanced-cleaner/
@@ -103,6 +102,7 @@ def anonymize_registrations(user: User):
         anonymize_registration(registration)
     # db.session has commit()
     db.session.commit()  # pylint: disable=no-member
+
 
 @cli.command("user")
 @click.argument("email", type=str)
--- a/indico_rock/plugins/anonymize/anonymize/cli.py
+++ b/indico_rock/plugins/anonymize/anonymize/cli.py
@@ -8,12 +8,11 @@ import click
 from hashlib import sha512
 from indico.cli.core import cli_group
 from indico.core.db import db
-from indico.modules.events.registration import Registration
+from indico.modules.events.registration.models.registrations import Registration
 from indico.modules.events.registration.models.form_fields import RegistrationFormField
 from indico.modules.users import User
 from indico.modules.users.util import search_users
 from indico.util.date_time import now_utc
-from indico.modules.events import User
 import re
 
 CLEAN_ATTRS_STR = ['affiliation', 'email', 'secondary_emails', ]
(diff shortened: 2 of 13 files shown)


indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -23,6 +25,16 @@
 def cli():
     """Anonymize users non-interactively."""
 
+# Extracted from:
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def _hash(val: str):
```
arturo-seijas, line 30: Can you add the return type hints?

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -72,6 +82,38 @@ def is_anonymized(user: User) -> bool:
         if not hash_regex.match(getattr(user, attr)):
             return False
 
+# Extracted from:
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def anonymize_registrations(user: User):
+    """Anonymize user by erasing registrations attributes.
```
arturo-seijas, line 88: s/registrations/registration?

indico_rock/rockcraft.yaml
```diff
@@ -59,7 +59,7 @@ parts:
         python3-saml \
         python-ldap \
         uwsgi \
-        ./autocreate
+        ./autocreate \
```
arturo-seijas, line 62: Can you order this two alphabetically?

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,163 @@
...
+
+    email = email.lower()
+
+    if email == "":
+        click.secho("E-mail should not be empty", fg="red")
+        ctx.exit(1)
+
+    users = User.query.filter(User.all_emails == email)
+    if not users.has_rows():
+        click.secho("User not found", fg="red")
```
arturo-seijas, line 136: With idempotence in mind, perhaps we should warn instead and return a 0 exit code

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,163 @@
...
+        include_pending=False,
+        include_blocked=False,
+        external=False,
+        allow_system_user=False,
+        email=email,
+    )
+
+    user = res.pop()
+
+    if not user.is_deleted or not is_anonymized(user):
```
arturo-seijas, line 159: Is this possible?

indico_rock/plugins/anonymize/setup.cfg
```diff
@@ -0,0 +1,28 @@
+[metadata]
+name = indico-plugin-anonymize
+version = 3.2
```
arturo-seijas, line 3: Does this value have to match the Indico version?

arturo-seijas (commented):
Some comments inline

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: It might be more than one registration I believe.

tests/integration/test_charm.py
```diff
@@ -132,6 +132,37 @@ async def test_add_admin(app: Application):
...
+    act: create an user and run the anonymize-user action
+    assert: check the output in the action result
+    """
+
+    # Application actually does have units
+    assert app.units[0]  # type: ignore
+
+    email = "anonymize@email.com"
+    # This is a test password
+    password = "somepassword"  # nosec
```
jdkandersson, line 147: What about using https://docs.python.org/3/library/secrets.html#secrets.token_hex?

tests/integration/test_charm.py
```diff
@@ -132,6 +132,37 @@ async def test_add_admin(app: Application):
...
 @pytest.mark.abort_on_fail
+async def test_anonymize_user(app: Application):
+    """
+    arrange: given charm in its initial state
+    act: create an user and run the anonymize-user action
+    assert: check the output in the action result
+    """
+
+    # Application actually does have units
+    assert app.units[0]  # type: ignore
```
jdkandersson, line 143: Perhaps put the unit either unit a fixture or into a local variable so that this type: ignore doesn't have to be done over and over?

@jnsgruk I think this is python-libjuju, is there anyone from the juju team I can tag to let them know about when we need to use type ignores? It would be good not to have to use type: ignore for libraries

tests/integration/test_charm.py
```diff
@@ -132,6 +132,37 @@ async def test_add_admin(app: Application):
...
+
+    email = "anonymize@email.com"
+    # This is a test password
+    password = "somepassword"  # nosec
+
+    # Application actually does have units
+    action_add_admin: juju.action.Action = await app.units[0].run_action(  # type: ignore
```
jdkandersson, line 150: What is the reason for creating an admin user here? Is it the most convenient way of crating a user we have? If so, perhaps worth it for a fixture?

tests/integration/test_charm.py
```diff
@@ -132,6 +132,37 @@ async def test_add_admin(app: Application):
...
+    assert action_add_admin.status == "completed"
+
+    # Application actually does have units
+    action_anonymize: juju.action.Action = await app.units[0].run_action(  # type: ignore
+        "anonymize-user", email=email
+    )
+    await action_anonymize.wait()
+    assert action_anonymize.status == "completed"
+    assert action_anonymize.results["user"] == email
+    assert f'User with email "{email}" correctly anonymized' in action_anonymize.results["output"]
```
jdkandersson, line 163: It is usually better just to check for key substrings because otherwise the error message is essentially written in two places, in the source code and in the test. In this case perhaps it would be enough to check that the email and `correctly anonymized` is in the output to establish that the code worked?

tests/unit/test_actions.py
```diff
@@ -203,3 +203,152 @@ def event_store_failure(arg):
...
+            environment=indico_env_config,
+        )
+
+        # Check if event results was properly set
+        mock_event.set_results.assert_called_with(
+            {"user": f"{email}", "output": (f"{email}", None)}
+        )
+
+    @patch.object(Container, "exec")
+    def test_anonymize_user_list(self, mock_exec):
```
jdkandersson, line 254: Consider unifying this and the previous test and parametrizing the email input (e.g., single, multiple and many)

tests/unit/test_actions.py
```diff
@@ -203,3 +203,152 @@ def event_store_failure(arg):
...
+    @patch.object(Container, "exec")
+    def test_anonymize_user_fail(self, mock_exec):
+        """
+        arrange: an email
+        act: when the _on_anonymize_user_action method is executed
+        assert: the indico command to anonymize the user is executed with the appropriate
+            parameters and the event results is set as expected
+        """
+        # Set Indico
+        charm: IndicoOperatorCharm = typing.cast(IndicoOperatorCharm, self.harness.charm)
```
jdkandersson, line 314: Looks like this is repeated in multiple tests, perhaps worth a fixture?

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: Agree, changed to exit 0 and fg yellow (warning)

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: This code was removed.

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,126 @@
...
+import re
+
+CLEAN_ATTRS_STR = ['affiliation', 'email', 'secondary_emails', ]
+CLEAN_ATTRS_SET= ['favorite_users', 'favorite_categories', 'identities']
+CLEAN_ATTRS_LIST = ['old_api_keys',]
+CLEAN_ATTRS_ANON = ['first_name', 'last_name', 'phone', 'address',]
+
+@cli_group(name="anonymize")
+def cli():
+    """Anonymize users non-interactively."""
```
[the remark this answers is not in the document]
amandahla: We're following the same pattern used by Indico team to add a command group. Example:
https://github.com/indico/indico-plugins-cern/blob/master/burotel/indico_burotel/cli.py

tests/integration/test_charm.py
amandahla: I tried but then we got the error "attr-defined"

actions.yaml
```diff
@@ -10,3 +10,9 @@ add-admin:
       type: string
       description: User password.
   required: [email, password]
+anonymize-user:
+  description: Anonymize user in Indico
```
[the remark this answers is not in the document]
amandahla: Changed.

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,126 @@
...
+# Extracted from:
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def anonymize_deleted_user(user: User):
+    """Anonymize user by erasing specific attributes.
+
+    Args:
+        user: Indico user
+    """
+
+    for attr in CLEAN_ATTRS_STR:
```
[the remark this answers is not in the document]
amandahla: Nice. Please verify if I get it right.

src/charm.py
```diff
@@ -971,6 +972,50 @@ def _add_admin_action(self, event: ActionEvent) -> None:
...
+        """Anonymize user in Indico.
+
+        Args:
+            event: Event triggered by the anonymize-user action
+        """
+        container = self.unit.get_container("indico")
+        indico_env_config = self._get_indico_env_config_str(container)
+
+        output_list = []
+        email_list_separator = ","
```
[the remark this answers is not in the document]
amandahla: Changed.

src/charm.py
```diff
@@ -971,6 +972,50 @@ def _add_admin_action(self, event: ActionEvent) -> None:
...
+    def _anonymize_user_action(self, event: ActionEvent) -> None:
+        """Anonymize user in Indico.
+
+        Args:
+            event: Event triggered by the anonymize-user action
+        """
+        container = self.unit.get_container("indico")
+        indico_env_config = self._get_indico_env_config_str(container)
+
+        output_list = []
```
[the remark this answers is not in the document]
amandahla: Changed. It will stop at first exception.

src/charm.py
```diff
@@ -971,6 +972,50 @@ def _add_admin_action(self, event: ActionEvent) -> None:
...
+                )
+                try:
+                    out = process.wait_output()
+                    output_list.append(out[0])
+                except ExecError as ex:
+                    logger.exception("Action anonymize-user failed: %s", ex.stdout)
+
+                    event.fail(
+                        # Parameter validation errors are printed to stdout
+                        f"Failed to anonymize user {event.params['email']}: {ex.stdout!r}"
```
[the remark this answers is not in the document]
amandahla: Same as above.

tests/integration/test_charm.py
amandahla: Changed to use add admin as fixture.

tests/integration/test_charm.py
amandahla: Good catch, changed.

tests/integration/test_charm.py
amandahla: Agree, changed.

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -15,57 +15,60 @@
 from indico.util.date_time import now_utc
 import re
 
-CLEAN_ATTRS_STR = ['affiliation', 'email', 'secondary_emails', ]
-CLEAN_ATTRS_SET= ['favorite_users', 'favorite_categories', 'identities']
-CLEAN_ATTRS_LIST = ['old_api_keys',]
-CLEAN_ATTRS_ANON = ['first_name', 'last_name', 'phone', 'address',]
+
+EMPTY_LIST = lambda _: list()
```
jdkandersson, line 19: I don't think the `lambda` is needed here, `EMPTY_LIST = list` should do the trick. Similar for `set`

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -15,57 +15,60 @@
...
+EMPTY_SET = lambda _: set()
+EMPTY_VAL = lambda _: ''
```
jdkandersson, line 21: I think this can be done using `EMPTY_VAL = str`

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -15,57 +15,60 @@
...
 
     Args:
-        user: Indico user
+        registration (Registration): User registration_
     """
-    registrations = Registration.query.filter(Registration.user_id == user.id)
-    for registration in registrations:
-        for fid, rdata in registration.data_by_field.items():
+    for fid, rdata in registration.data_by_field.items():
             fieldtype = RegistrationFormField.get(oid=fid).input_type
```
jdkandersson, line 72: Looks like this can be un-indented by 4 spaces now that the nesting is reduced?

src/charm.py
```diff
@@ -972,18 +975,21 @@ def _add_admin_action(self, event: ActionEvent) -> None:
                     f"Failed to create admin {event.params['email']}: {ex.stdout!r}"
                 )
 
-    def _anonymize_user_action(self, event: ActionEvent) -> None:
-        """Anonymize user in Indico.
+    def _execute_anonymize_cmd(self, event: ActionEvent) -> Generator:
```
jdkandersson, line 978: I think the `Iterator[<type>]` should be ok here, the `Generator` type is needlessly complex. I'm not sure what `<type>` should be, whatever `out[0]` is, maybe a `str`?

src/charm.py
```diff
@@ -1000,21 +1006,42 @@ def _anonymize_user_action(self, event: ActionEvent) -> None:
...
+
+        Args:
+            event: Event triggered by the anonymize-user action
+        """
+        if len(event.params["email"].split(EMAIL_LIST_SEPARATOR)) > EMAIL_LIST_MAX:
+            fail_msg = "List of more than 50 emails are not allowed"
+            logger.error("Action anonymize-user failed: %s", fail_msg)
+            event.fail(fail_msg)
+            return
+        output_list = []
```
jdkandersson, line 1028: Is this assignment still needed?

src/charm.py
```diff
@@ -1000,21 +1006,42 @@ def _anonymize_user_action(self, event: ActionEvent) -> None:
...
+            output_list.append(fail_msg)
+            event.fail(
+                # Parameter validation errors are printed to stdout
+                fail_msg
+            )
+        finally:
+            event.set_results(
+                {
+                    "user": f"{event.params['email']}",
+                    "output": (EMAIL_LIST_SEPARATOR.join(output_list), None),
```
jdkandersson, line 1042: I think what will happen here is that either `output_list` will contain all the emails that succeeded or it will just contain the `fail_message`, is that as intended?

tests/integration/test_charm.py
```diff
@@ -116,7 +119,7 @@ async def test_add_admin(app: Application):
     # Application actually does have units
     assert app.units[0]  # type: ignore
 
-    email = "sample@email.com"
+    email = ADMIN_USER_EMAIL
     # This is a test password
     password = "somepassword"  # nosec
```
jdkandersson, line 124: Should we change this to https://docs.python.org/3/library/secrets.html#secrets.token_hex? That way we resolve a long hanging fruit issue

tests/integration/test_charm.py
jdkandersson: Could you share the code that caused this please?

src/charm.py
amandahla: Yes, because I'm using inside the except too.

src/charm.py
amandahla: I was considering that after raising an Exception, the output_list would contain the previous results but that's not the case so I changed a little to get all the results until an Exception is raised.

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: What an eye, good catch!

src/charm.py
amandahla: Yes, is a str. Changed.

src/charm.py
amandahla: Same answer: "I was considering that after raising an Exception, the output_list would contain the previous results but that's not the case so I changed a little to get all the results until an Exception is raised."

tests/integration/test_charm.py
amandahla: Yes, changed.

amandahla:
Answering your comment:

Fixed by
```
assert hasattr(app, "units")
assert app.units[0]
```

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: I changed for [] because of "R1734: Consider using [] instead of list() (use-list-literal)"

tests/unit/_patched_charm.py
```diff
@@ -62,3 +62,5 @@ def stop(self):
 
 pgsql_patch = _PGSQLPatch()
 IndicoOperatorCharm = __import__("charm").IndicoOperatorCharm
+EMAIL_LIST_MAX = __import__("charm").EMAIL_LIST_MAX
```
jdkandersson, line 65: This seems unusual, would `from charm import EMAIL_LIST_MAX` work? Similar for the others?

tests/unit/test_actions.py
```diff
@@ -310,14 +328,10 @@ def test_anonymize_user_fail(self, mock_exec):
...
-        # Set Indico
-        charm: IndicoOperatorCharm = typing.cast(IndicoOperatorCharm, self.harness.charm)
-        charm._get_installed_plugins = MagicMock(return_value="")
-        charm._get_indico_secret_key_from_relation = MagicMock(return_value="")
-        self.harness.container_pebble_ready("indico")
+        charm = self._set_indico()
 
         # Set Mock
-        email = "sample@email.com"
+        email = token_hex(16)
```
jdkandersson, line 334: There is a nice library that can generate values for us: https://factoryboy.readthedocs.io/en/stable/

tox.ini
```diff
@@ -101,12 +102,13 @@ commands =
...
       --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg \
       --skip {toxinidir}/indico_rock/plugins/autocreate/.mypy_cache
+      --skip {toxinidir}/indico_rock/plugins/anonymize/.mypy_cache
     # pflake8 wrapper supports config from pyproject.toml
     pflake8 {[vars]plugins_path} --ignore=W503
     isort --check-only --diff {[vars]plugins_path}
     black --check --diff {[vars]plugins_path}
     mypy {[vars]plugins_path}
-    pylint {[vars]plugins_path} --ignore-paths {[vars]plugins_path}/autocreate/build
+    pylint {[vars]plugins_path} --ignore-paths indico_rock/plugins/autocreate/build/,indico_rock/plugins/anonymize/build/
```
jdkandersson, line 111: Any particular reason for no longer using the variable please?

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -4,40 +4,33 @@
...
-    'last_name': HASH_VAL,
-    'phone': HASH_VAL,
-    'address': HASH_VAL,
+    "affiliation": str(),
+    "email": str(),
+    "secondary_emails": str(),
+    "favorite_users": set(),
+    "favorite_categories": set(),
+    "identities": set(),
+    "old_api_keys": [],
```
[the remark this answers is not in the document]
amandahla: I believe it won't make much of a difference since we only need an empty value to update the Indico database as expected.

tests/unit/_patched_charm.py
amandahla: It was my first try but there is a comment warning: "This patch needs to be run prior to the importing of the main module since the main module uses ``ops.lib.use``". So I did this instead.

src/charm.py
```diff
@@ -1021,27 +1024,18 @@ def _anonymize_user_action(self, event: ActionEvent) -> None:
...
-                fail_msg
-            )
-        finally:
-            event.set_results(
-                {
-                    "user": f"{event.params['email']}",
-                    "output": (EMAIL_LIST_SEPARATOR.join(output_list), None),
-                }
-            )
+        output_list = list(self._execute_anonymize_cmd(event))
```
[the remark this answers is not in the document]
amandahla: Since the _execute_anonymize_cmd is handling the ExecError Exception, it needs the event parameter to set the fail message (event.fail) .

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -48,7 +41,8 @@ def _hash(val: str) -> str:
     Returns:
         First 12 characters from encoded data in hexadecimal format
     """
-    return sha512(val.encode('utf-8')+now_utc().isoformat().encode('utf-8')).hexdigest()[:12]
+    return sha512(val.encode("utf-8") + now_utc().isoformat().encode("utf-8")).hexdigest()[:12]
```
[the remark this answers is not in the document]
amandahla: When I changed this mypy complains with " error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior  [str-bytes-safe]"
I've tried many combinations without success.

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -60,7 +54,11 @@ def anonymize_deleted_user(user: User):
     """
     for attr, anonymize_val in CLEAN_ATTRS.items():
         current_val = getattr(user, attr)
-        setattr(user, attr, anonymize_val(current_val))
+        setattr(user, attr, anonymize_val)
```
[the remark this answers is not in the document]
amandahla: Same comment above.

tests/unit/test_actions.py
amandahla: Nice :) I suggest doing this in a new PR so we could change all the tests to use this.

tox.ini
amandahla: I believe it's something related to this [issue](https://github.com/PyCQA/pylint/issues/6964) When I added a new directory to ignore-paths using the variable (that is an absolute path) is ignored. But setting both as relative paths then it works.

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
jdkandersson: Sharing mutable values, even if we don't think it will be a problem in a particular case, is a way to open us up to a lot of weird and unexpected bugs. It is better not to get into the habit of doing this

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
jdkandersson: Now that I'm looking at this, what is the intent? I'm guessing this is supposed to anonymize data? Why use a hashing algorithm for that please?

src/charm.py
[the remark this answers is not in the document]
jdkandersson: Hmm, I think that is also quite coupled. How about `self._execute_anonymize_cmd` takes the emails to process and yields both the message and whether processing the email succeeded? Then this function can inspect the results and set `event.fail` accordingly?

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
amandahla: I'm afraid that is not documented but I would say that this is because Indico wouldn't work/allow an empty value for this field.
Source: https://github.com/bpedersen2/indico-cron-advanced-cleaner/blob/master/indico_cron_advanced_cleaner/tasks.py#L60

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
amandahla: Marking this as resolved so we can continue/finish in the comment above.

src/charm.py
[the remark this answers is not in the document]
amandahla: I see your point but I think I would prefer this way so we don't need to check each status. When an expectation is raised, the generator just stops, set the event.fail accordingly and the action finishes.
For now, I would keep with this approach that could work with different parameters as well and we can change it in the future if needed.

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,142 @@
...
+# https://github.com/bpedersen2/indico-cron-advanced-cleaner/
+def _hash(val: str) -> str:
+    """Create hash of val.
+
+    Args:
+        val: The string from the hash should be created
+    Returns:
+        First 12 characters from encoded data in hexadecimal format
+    """
+    return sha512(val.encode("utf-8") + now_utc().isoformat().encode("utf-8")).hexdigest()[:12]
```
weiiwang01, line 44: I am a little curious about the purpose of using hash here. It's not deterministic since it uses the current time as salt. And it's not very secure, since the salt entropy is not very high.

I think this can be replaced with random tokens:
```
import secrets

secrets.token_hex(8)
```

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,142 @@
...
+        elif fieldtype == "country":
+            # Keep country for now
+            pass
+        # other field types are not touched (choice, multiple choice, radio)
+
+    registration.first_name = _hash(registration.first_name)
+    registration.last_name = _hash(registration.last_name)
+    registration.email = _hash(registration.email)
+    if registration.user:
+        registration.user = None
```
weiiwang01, line 89: Update user information with random string can be confusing. If another user access the anonymized page, the user might be confused by the random string inside that page.

Maybe we can update fields to a more understandable placeholder like "REMOVED"?

src/charm.py
```diff
@@ -971,6 +975,70 @@ def _add_admin_action(self, event: ActionEvent) -> None:
...
+        indico_env_config = self._get_indico_env_config_str(container)
+        for email in event.params["email"].split(EMAIL_LIST_SEPARATOR):
+            cmd = [
+                "/usr/local/bin/indico",
+                "anonymize",
+                "user",
+                email,
+            ]
+
+            if container.can_connect():
```
weiiwang01, line 1000: I think it might be the best to return error when the container is not ready.

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: @weiiwang01 @jdkandersson Since you both questioned that, I opened an issue [here](https://github.com/bpedersen2/indico-cron-advanced-cleaner/issues/2) to find out if there is a specific reason for doing that. 

I read that hash is one of the techniques that you can use for pseudonymization but I don't have enough knowledge regarding GDPR to say if this is enough.

https://korniichuk.medium.com/gdpr-guide-2-7c399b44ba3#adce

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
amandahla: I'll mark this one as resolved so we can discuss it in the one opened by Weii, just to centralize the discussion.

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: Same answer as above. I'll mark this one as resolved so we can centralize the discussion regarding how to anonymize personal data (hash usage or not) there.

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
jdkandersson: It would be worth investigating that, you also hav to call the constructor so it isn't sufficient just to changing the dictionary, the function setting those values also needs to be changed to call the constructor

indico_rock/plugins/anonymize/anonymize/cli.py
jdkandersson: Why not just replace it with a random value if there has to be a value or just make it empty? It would be simpler and would be much less likely to be an issue under GDPR (although I'm not an expert in it either)

jdkandersson:
@gregory-schiano this is a good PR to take a look at

indico_rock/plugins/anonymize/anonymize/cli.py
arturo-seijas: I'd go replacing with a random value. Once anonymized, the user data should be non recoverable. On the other hand, empty values might cause Indico to crash

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if email == "":
+        click.secho("E-mail should not be empty", fg="red")
+        ctx.exit(1)
+
+    users = User.query.filter(User.all_emails == email)
+    if not users.has_rows():
+        click.secho(f"No user found for email {email}", fg="yellow")
+        ctx.exit(0)
+    user = users.first()
```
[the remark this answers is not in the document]
amandahla: I didn't find it in the docs but I'm assuming from [here](https://github.com/indico/indico/blob/master/indico/modules/auth/forms.py#L26) that an email is unique. There are multiple emails for a user but there are no multiple users with the same e-mail.

Note: email is not the id. The User entity has an [exclusive field](https://github.com/indico/indico/blob/master/indico/modules/users/models/users.py#L195) for that.

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: - I believe empty is not the best option because Indico could crash while rendering forms/lists.
- The hash is not user-friendly for visible fields.
- Emails: From [here](https://github.com/indico/indico/blob/master/indico/modules/users/models/emails.py#L15), I believe email should be unique so while replacing we must find a way to keep this as a distinct value.

indico_rock/plugins/anonymize/anonymize/cli.py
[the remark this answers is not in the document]
amandahla: Thanks for the explanation. Changed.

src/charm.py
amandahla: Good point!

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: @gregory-schiano I applied your suggestion(uuid) now, thanks.

src/charm.py
```diff
@@ -997,22 +997,27 @@ def _execute_anonymize_cmd(self, event: ActionEvent) -> Iterator[str]:
...
             ]
 
-            if container.can_connect():
-                process = container.exec(
-                    cmd,
-                    user="indico",
-                    working_dir="/srv/indico",
-                    environment=indico_env_config,
+            if not container.can_connect():
+                logger.error(
```
arturo-seijas, line 1001: Can you set a failed status for the action with the appropriate message instead of just logging the error?

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # We mark as deleted so won't appear in search users forms
+    user.is_deleted = True
+    anonymize_deleted_user(user)
+    # Anonymize registrations
+    anonymize_registrations(user)
```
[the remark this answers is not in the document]
amandahla: Changed.

indico_rock/plugins/anonymize/anonymize/cli.py
jdkandersson: This is a better approach than hash

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -53,11 +48,9 @@ def anonymize_deleted_user(user: User):
         user: Indico user
     """
     for attr, anonymize_val in CLEAN_ATTRS.items():
-        current_val = getattr(user, attr)
-        setattr(user, attr, anonymize_val)
+        setattr(user, attr, anonymize_val())
     for attr in HASH_ATTRS:
```
jdkandersson, line 52: Should this still be called `HASH_ATTRS`? Also, I think `CLEAN_ATTRS` could be updated with the values from `HASH_ATTRS` and just have the `_generate_uuid` function assigned, then all the attributes are nicely handled

indico_rock/plugins/anonymize/anonymize/cli.py
```diff
@@ -82,9 +78,9 @@ def anonymize_registration(registration: Registration):
             pass
         # other field types are not touched (choice, multiple choice, radio)
 
-    registration.first_name = _hash(registration.first_name)
-    registration.last_name = _hash(registration.last_name)
-    registration.email = _hash(registration.email)
+    registration.first_name = first_name
```
jdkandersson, line 81: Should this also be replaced by a uuid?

src/charm.py
```diff
@@ -997,22 +997,30 @@ def _execute_anonymize_cmd(self, event: ActionEvent) -> Iterator[str]:
...
+            )
+            try:
+                out = process.wait_output()
+                yield out[0].replace("\n", "")
+            except ExecError as ex:
+                logger.exception("Action anonymize-user failed: %s", ex.stdout)
+                fail_msg = f"Failed to anonymize user {event.params['email']}: {ex.stdout!r}"
+                event.fail(fail_msg)
+                yield fail_msg
+                return
```
jdkandersson, line 1023: Something to consider: just keep going so that at least all the users without an error get processed. Currently, there is an arbitrary cut off depending on how users are listed that all the users *before* the first failure are processed and none of the emails *after` the first failure are processed. I think just deleting the `return` should do the trick

jdkandersson:
There you go, this is the prize for getting to over 100 interactions: 🎆

gregory-schiano:
I'm fine with this PR, I'll approve once the last comments are managed (to avoid my approval being canceled by a new commit 😛 )

src/charm.py
amandahla: In that case, I still should set the event.fail, right? Maybe with a different message like "Failed to anonymize one or more users. Please, verify the results."

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: I believe so.
It was replaced by uuid here:
```
first_name = _generate_uuid()
email = f"{first_name}@{_generate_uuid()}.local"
```

arturo-seijas (dismissed):
LGTM

indico_rock/plugins/anonymize/anonymize/cli.py
amandahla: Nice! Changed.

jdkandersson (approved):

src/charm.py
jdkandersson: Yeh, I wonder what happens in that case, could be worth an integration test where one of the middle users fails, up to you

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

src/charm.py
amandahla: It's a nice idea, I'll create a new story for that because I believe it would be challenging to force/simulate an error in this situation.

PR #249, opened 2023-03-09
Add SAML integration test

gtrkiller:
This PR adds an integration test to test the whole SAML authentication process.


.github/workflows/integration_test_with_secrets
```diff
@@ -0,0 +1,45 @@
...
+          docker build -f indico-nginx.Dockerfile -t localhost:32000/indico-nginx:test .
+          docker push localhost:32000/indico-nginx:test
+      - name: Run integration tests
+        run: |
+          tox -e integration -- \
+          -m requires_secrets \
+          --saml-email ${{ secrets.TEST_SAML_EMAIL }} \
+          --saml-password ${{ secrets.TEST_SAML_PASSWORD }} \
+          --indico-nginx-image=localhost:32000/indico-nginx:test \
+          --indico-image=localhost:32000/indico:test
```
arturo-seijas, line 45: Missing blank line at EOF

tests/integration/test_charm.py
```diff
@@ -118,3 +128,80 @@ async def test_add_admin(app: Application):
...
+@pytest.mark.abort_on_fail
+@pytest.mark.requires_secrets
+async def test_saml_auth(app: Application, saml_email: str, saml_password: str):
+    """
+    arrange: given charm in its initial state
+    act: configure a SAML target url and fire SAML authentication
+    assert: The SAML authentication process is executed successfully.
+    """
+    await app.set_config({"site_url": "https://indico.local"}),  # type: ignore[attr-defined] # pylint: disable=W0106 # noqa
+    await app.set_config({"saml_target_url": STAGING_UBUNTU_SAML_URL}),  # type: ignore[attr-defined] # pylint: disable=expression-not-assigned # noqa
```
weiiwang01, line 143: I think this two can be merged into one line:

```
    await app.set_config({
        "site_url": "https://indico.local",
        "saml_target_url": STAGING_UBUNTU_SAML_URL
    })
```

tests/integration/test_charm.py
```diff
@@ -118,3 +128,80 @@ async def test_add_admin(app: Application):
...
+            return original_getaddrinfo("127.0.0.1", *args[1:])
+        return original_getaddrinfo(*args)
+
+    with patch.multiple(socket, getaddrinfo=patched_getaddrinfo):
+        session = requests.session()
+        dashboard_page = session.get(
+            f"https://{host}/user/dashboard/", verify=False, allow_redirects=False
+        )
+        assert dashboard_page.status_code == 302
+        print(dashboard_page.headers["Location"])
```
weiiwang01, line 160: Should this debug statement be removed?

tests/integration/test_charm.py
```diff
@@ -118,3 +128,80 @@ async def test_add_admin(app: Application):
...
+                "user-intentions": "login",
+                "password": saml_password,
+                "next": "/saml/process",
+                "continue": "",
+                "openid.usernamesecret": "",
+                "RelayState": "indico.local",
+            },
+            headers={"Referer": login_page.url},
+        )
+        print(saml_callback.text)
```
weiiwang01, line 185: There might be some sensitive information in the small callback page, so I think print the content here might be a little bit unnecessary.

tests/integration/test_charm.py
gtrkiller: yes, I'll remove it. Thank you for noticing!

tox.ini
```diff
@@ -97,7 +97,7 @@ deps =
     types-requests
     -r{toxinidir}/requirements.txt
 commands =
-    codespell {[vars]plugins_path} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
+    codespell -L astroid {[vars]plugins_path} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
```
arturo-seijas, line 100: This has been fixed in the operator workflows. You don't need to add this exception anymore

arturo-seijas (commented):
Some comments inline

.github/workflows/integration_test_with_secrets.yaml
```diff
@@ -0,0 +1,52 @@
...
+          sudo microk8s enable hostpath-storage ingress registry
+          sudo microk8s kubectl -n kube-system rollout status -w deployment/hostpath-provisioner
+          sudo microk8s kubectl -n ingress rollout status -w daemonset.apps/nginx-ingress-microk8s-controller
+          sudo microk8s kubectl -n container-registry rollout status -w deployment/registry
+      - name: Install tox
+        run: python3 -m pip install tox
+
+      - name: Install rockcraft
+        run: sudo snap install --edge rockcraft --classic
+
```
yhaliaw, line 29: Sometimes there are blank lines between steps but sometimes there is not. Is there a reason behind this? 
Thanks.

yhaliaw (dismissed):
LGTM

tests/integration/test_charm.py
```diff
@@ -118,3 +128,99 @@ async def test_add_admin(app: Application):
...
+
+    host = "indico.local"
+    original_getaddrinfo = socket.getaddrinfo
+
+    def patched_getaddrinfo(*args):
+        if args[0] == host:
+            return original_getaddrinfo("127.0.0.1", *args[1:])
+        return original_getaddrinfo(*args)
+
+    with patch.multiple(socket, getaddrinfo=patched_getaddrinfo):
```
yanksyoon, line 165:
```suggestion
with patch.multiple(socket, getaddrinfo=patched_getaddrinfo), requests.session() as session:
```
Would this syntax be more readable?

tox.ini
```diff
@@ -131,6 +131,7 @@ commands =
 [testenv:static]
 description = Run static analysis tests
 deps =
+    bandit[toml]
```
yanksyoon, line 134: Correct me if I'm wrong, but I think a single bandit[toml] might be enough :)

yanksyoon (dismissed):
Minor suggestions, LGTM!

src/charm.py
```diff
@@ -593,38 +594,79 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
-                "sp": {
-                    "entityId": self.config["site_url"],
-                },
-                "idp": {
-                    "entityId": "https://login.ubuntu.com",
-                    "singleSignOnService": {
-                        "url": "https://login.ubuntu.com/saml/",
-                        "binding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
+            saml_config = {}
+            if self.config["saml_target_url"] == UBUNTU_SAML_URL:
```
jdkandersson, line 598: Do you think it would be better to read these from a, for example, JSON file?

src/charm.py
```diff
@@ -593,38 +594,79 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
-                        "EwJHQjETMBEGA1UECBMKU29tZS1TdGF0ZTEWMBQGA1UEChMNQ2Fub25pY2FsIEx0ZIIJALNN/"
-                        "vxaR1hyMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEFBQADgYEArTGbZ1rg++aBxnNuJ7eho6"
-                        "2JKKtRW5O+kMBvBLWi7fKck5uXDE6d7Jv6hUy/gwUZV7r5kuPwRlw3Pu6AX4R60UsQuVG1/VV"
-                        "VI7nu32iCkXx5Vzq446IkVRdk/QOda1dRyq0oaifUUhJfwVFSsm95ENDFdGqD0raj7g77ajcB"
-                        "Mf8="
-                    ),
-                },
-            }
+                }
+            elif self.config["saml_target_url"] == STAGING_UBUNTU_SAML_URL:
```
jdkandersson, line 631: Would this be better as configuration? Usually the source code should not include environment differences since it is only applicable to a given environment. Perhaps the differences between staging and production should be configuration?

tests/integration/conftest.py
```diff
@@ -83,6 +107,6 @@ async def app(
         ops_test.model.add_relation(app_name, "redis-cache"),
         ops_test.model.add_relation(app_name, "nginx-ingress-integrator"),
     )
-    await ops_test.model.wait_for_idle(status="active")
+    await ops_test.model.wait_for_idle(status="active", raise_on_error=False)
```
jdkandersson, line 110: What is the reason for not raising if there is an error please?

tests/integration/test_charm.py
```diff
@@ -118,3 +128,99 @@ async def test_add_admin(app: Application):
...
+    saml_email: str,
+    saml_password: str,
+    requests_timeout: float,
+):
+    """
+    arrange: given charm in its initial state
+    act: configure a SAML target url and fire SAML authentication
+    assert: The SAML authentication process is executed successfully.
+    """
+    await app.set_config(  # type: ignore[attr-defined] # pylint: disable=W0106
```
jdkandersson, line 148: We have a standard for including a description for why linting rules are disabled, is that applicable here?

jdkandersson: Similar comment for any other disable rules

src/charm.py
gtrkiller: I think so. It is included in the ISD024 spec as one of the possibilities, but will keep it this way until that spec advances a bit.

tests/integration/conftest.py
gtrkiller: The postgresql-k8s charm is really flaky on deploy, so it has a high probability of raising an error caused by a hook and then correcting itself while there is nothing wrong and the deployment ends successfully. If I don't add this we may have to execute the tests up to 4 times more to avoid this flaky mistake by that charm.

src/charm.py
gtrkiller: See above

tox.ini
```diff
@@ -132,7 +132,6 @@ commands =
 description = Run static analysis tests
 deps =
     bandit[toml]
-    bandit
     toml
```
arturo-seijas, line 135: You don't need toml either

arturo-seijas (approved):
LGTM

yanksyoon (approved):
LGTM!

src/charm.py
jdkandersson: If we have a spec on this that isn't finalised yet, what is the reason for merging this code into main? Shouldn't we wait for the spec to be reviewed?

tests/integration/conftest.py
jdkandersson: @yanksyoon FYI, perhaps also applicable for the issues we're having with `upload-charm-docs`?

yanksyoon: What i've seen happen, in other times though, is that when doing `raise_on_error=False` the test would continue while the unit is in error causing the test to fail.
I've also tried [wrapping the tests](https://github.com/canonical/wordpress-k8s-operator/pull/57/commits/a9a873f733fded0bb97b87580469ba9ee6824f4c) with try-except blocks but didn't quite get it to work either.
Let me try this method on the postgresql-k8s for upload-charm-docs, maybe the behavior might be different due to different nature of error than loki-k8s.
Thanks for the heads up!

jdkandersson: @delgod FYI, it looks like the postgres-k8s charm is having stability issues?

PR #253, opened 2023-04-04
Fix nginx rock build

arturo-seijas:
Changed as discussed in https://github.com/canonical/rockcraft/issues/225

### DIFF
--- a/nginx_rock/rockcraft.yaml
+++ b/nginx_rock/rockcraft.yaml
@@ -8,29 +8,6 @@ license: Apache-2.0
 platforms:
   amd64:
 parts:
-  add-user:
-    plugin: nil
-    overlay-script: |
-      mkdir $CRAFT_OVERLAY/etc
-      chmod 755 $CRAFT_OVERLAY/etc
-      groupadd -R $CRAFT_OVERLAY --gid 2000 nginx
-      useradd -R $CRAFT_OVERLAY --system --gid 2000 --uid 2000 --no-create-home nginx
-  nginx-conf:
-    plugin: dump
-    source: etc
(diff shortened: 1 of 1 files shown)


gtrkiller (dismissed):
LGTM

jdkandersson (dismissed):

nginx_rock/rockcraft.yaml
```diff
@@ -45,3 +42,7 @@ parts:
       lib/python3.10/site-packages/indico/web/static: srv/indico/static
     stage:
       - srv/indico/static
+    stage-packages:
+      - build-essential
```
weiiwang01, line 46: Is the `build-essential` required in the OCI image, that package is quite large.

gregory-schiano: I think that since there's a stage entry, it will only stage the specified files in the stage entry, the packages are just installed during this phase

gregory-schiano (approved):
LGTM

nginx_rock/rockcraft.yaml
arturo-seijas: Exactly. And yes, I'm afraid it's needed

amandahla (approved):
LGTM

nginx_rock/rockcraft.yaml
weiiwang01: 👍

weiiwang01 (approved):

arturo-seijas:
Load tests are failing but if there's an issue there it will be addressed as part of https://github.com/canonical/indico-operator/pull/252

PR #267, opened 2023-05-12
Select channel track based on version

arturo-seijas:
Select channel track based on the version of Indico present in the rock file


gtrkiller (approved):
LGTM

yanksyoon (approved):
LGTM

PR #280, opened 2023-06-09
Fix publish workflow

arturo-seijas:
Follow up #279


mthaddon (approved):

gregory-schiano (approved):
LGTM
Integration tests probably requires a rerun

PR #287, opened 2023-07-21
Adapt integration tests to reuse the packed charm

arturo-seijas:
Adapt the integration tests to reuse the exisintg charm artifact as per https://github.com/canonical/operator-workflows/pull/162
Load test are disabled due to their flakiness. There's a scheduled weekly execution already

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -12,7 +12,7 @@ jobs:
       chaos-enabled: true
       chaos-experiments: pod-delete
       extra-arguments: -m "not (requires_secrets)"
-      load-test-enabled: true
+      load-test-enabled: false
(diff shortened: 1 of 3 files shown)


gtrkiller (approved):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       chaos-enabled: true
       chaos-experiments: pod-delete
       extra-arguments: -m "not (requires_secrets)"
-      load-test-enabled: true
+      load-test-enabled: false
```
mthaddon, line 15: We're disabling load tests? That's not mentioned in the description of the PR - can you update it to include a comment about that and explain why?

arturo-seijas: Done

jdkandersson (approved):

PR #291, opened 2023-07-26
Install Prometheus exporters along with the application

gtrkiller:
Delete the exporter container and install them as snaps in the Indico image.


pyproject.toml
```diff
@@ -54,4 +54,4 @@ markers = [
 ]
 
 [tool.pylint]
-disable = "wrong-import-order"
+disable = "wrong-import-order, duplicate-code"
```
jdkandersson, line 57: What is the reason for needing this please?

src/charm.py
```diff
@@ -271,7 +262,16 @@ def _config_pebble(self, container: Container) -> None:
         pebble_config = pebble_config_func(container)
         container.add_layer(container.name, pebble_config, combine=True)
         if container.name == "indico":
+            for item in ["statsd", "celery"]:
```
jdkandersson, line 265: Should this be a tuple?

jdkandersson: Also, since there are 2 items, is the `for` loop worth it? It is quite complex compared to the 4 lines of code that are needed?

merkata (dismissed):
Looks good, apart of what David commented on, nothing to add.

tests/unit/test_core.py
```diff
@@ -451,6 +451,7 @@ def test_config_changed_when_saml_target_url_invalid(self, mock_exec):
         self.set_up_all_relations()
         self.harness.set_leader(True)
 
+        # pylint: disable=duplicate-code
```
jdkandersson, line 454: What specifically is pylint complaining about being a duplicate here? In tests it means there is potential for a fixture

jdkandersson: perhaps worth making it into a fixture now?

amandahla (dismissed):
LGTM :)

Just wanted to see if the images are too much bigger or not than before.

amandahla (approved):

jdkandersson (approved):

PR #298, opened 2023-08-14
Use SAML groups identity provider

cbartz:
Use the identity provider from the [Flask-Multipass-SAML-Groups](https://github.com/canonical/flask-multipass-saml-groups) plugin if it has been configured.

### DIFF
--- a/docs/reference/plugins.md
+++ b/docs/reference/plugins.md
@@ -3,3 +3,7 @@
 Indico provides many plugins that can be used to extend functionalities without the need to change the Indico core. See more information in [Extending Indico with plugins](https://docs.getindico.io/en/latest/plugins/).
 
 You can configure additional plugins using the [`external_plugins`](https://charmhub.io/indico/configure#external_plugins) configuration option.
+
+There is a special treatment for the [Flask-Multipass-SAML-Groups](https://github.com/canonical/flask-multipass-saml-groups/) plugin: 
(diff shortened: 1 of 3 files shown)


arturo-seijas (approved):
LGTM

mthaddon (approved):
Change looks good. There seem to be two unrelated test failures here, need to confirm if it's expected that a change to the docs proposed in a PR would fail tests (i.e. changes need to be applied before landing?), and I see a branch from Arturo in progress to address SAML integration test failures, but I'm not sure it's the same issue as what you're seeing here...

arturo-seijas:
> Change looks good. There seem to be two unrelated test failures here, need to confirm if it's expected that a change to the docs proposed in a PR would fail tests (i.e. changes need to be applied before landing?), and I see a branch from Arturo in progress to address SAML integration test failures, but I'm not sure it's the same issue as what you're seeing here...

Let's retry. The runner run out of disk

cbartz:
> Change looks good. There seem to be two unrelated test failures here, need to confirm if it's expected that a change to the docs proposed in a PR would fail tests (i.e. changes need to be applied before landing?), and I see a branch from Arturo in progress to address SAML integration test failures, but I'm not sure it's the same issue as what you're seeing here...

There was a diff in the base of `docs/reference/plugins.md` and the discourse post because the discourse post had no trailing newline and the source did. I fixed it by recreating the `upload-charm-docs/base-content` tag and updating the discussion post.

PR #320, opened 2023-09-06
Migrate to new db relation library

arturo-seijas:
Applicable spec: N/A

### Overview

Migrate to new db relation library

### Rationale

The old relation mechanism is deprecated in favor of https://charmhub.io/data-platform-libs/libraries/data_interfaces

### Juju Events Changes

Removed old events and added the new equivalent ones

### Module Changes

Added a database observer module

### Library Changes

Added the data_interfaces lib

### Checklist


.github/workflows/integration_test.yaml
```diff
@@ -21,3 +21,4 @@ jobs:
       zap-rules-file-name: "zap_rules.tsv"
       trivy-fs-enabled: true
       trivy-image-config: "trivy.yaml"
+      tmate-debug: true
```
jdkandersson, line 24: I'm guessing this should be removed?

jdkandersson:
I think this PR is partially compliant with ISD054 - e.g., the charm state is not implemented. Not saying we should implement CharmState now - just note that the PR is not fully compliant for now and consider a refactor of the charm for the backlog. It is good that a database observer is implemented

src/database_observer.py
```diff
@@ -0,0 +1,72 @@
...
+    def _on_database_created(self, _: DatabaseCreatedEvent) -> None:
+        """Handle database created."""
+        self._charm.on.config_changed.emit()
+
+    def _on_endpoints_changed(self, _: DatabaseEndpointsChangedEvent) -> None:
+        """Handle endpoints change."""
+        self._charm.on.config_changed.emit()
+
+    @property
+    def uri(self) -> typing.Optional[str]:
```
jdkandersson, line 55: Should this be a function or a property? It looks a bit more complex and also consider whether there could be an exception that gets raised

tests/unit/test_database_observer.py
```diff
@@ -0,0 +1,107 @@
...
+  database:
+    interface: postgresql_client
+"""
+
+
+class ObservedCharm(ops.CharmBase):
+    """Class for requirer charm testing."""
+
+    def __init__(self, *args):
+        """Init method for the class.
```
jdkandersson, line 23: Perhaps `Construct database observer`?

jdkandersson: Or even just `Construct.`

tests/unit/test_database_observer.py
```diff
@@ -0,0 +1,107 @@
...
+        self.events.append(event)
+
+
+def test_database_created_emits_config_changed_event():
+    """
+    arrange: set up a charm and a database relation.
+    act: trigger a database created event.
+    assert: a config change event is emitted.
+    """
+    harness = Harness(ObservedCharm, meta=REQUIRER_METADATA)
```
jdkandersson, line 48: Please make sure the tests follow this standard: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

jdkandersson: specifically around an empty line between arrange, act and assert code blocks

src/database_observer.py
arturo-seijas: > Should this be a function or a property?
Do you see any reason why this shouldn't be a property? It might be a bit complex, but conceptually, I'd say it qualifies as such

src/charm.py
```diff
@@ -230,8 +185,7 @@ def _are_relations_ready(self, _) -> bool:
         if self._get_redis_cache_rel() is None:
             self.unit.status = WaitingStatus("Waiting for redis-cache availability")
             return False
-        # mypy misinterprets this line, reports something about function overloading
-        if not self._stored.db_uri:  # type: ignore
+        if not self.database.uri:
```
amandahla, line 188: Make no difference but just to keep the pattern:
```
        if self._get_redis_broker_rel() is None:
            self.unit.status = WaitingStatus("Waiting for redis-broker availability")
            return False
        if self._get_redis_cache_rel() is None:
            self.unit.status = WaitingStatus("Waiting for redis-cache availability")
            return False
        if self.database.uri is None:
            self.unit.status = WaitingStatus("Waiting for database availability")
            return False
        return True
```

tests/unit/test_database_observer.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    assert harness.charm.database.uri == (
+        "postgresql://user1:somepass@postgresql-k8s-primary.local:5432/indico"
+    )
+
+
+def test_uri_when_no_relation_data():
+    """
+    arrange: set up a charm and a database relation with an empty databag.
```
amandahla, line 105: Out of curiosity: When this would happen? relation broken?

arturo-seijas: On creation, the relation data is not populated, iirc

amandahla (dismissed):
LGTM

tests/integration/conftest.py
```diff
@@ -61,8 +61,8 @@ async def app(
     assert ops_test.model
     # Deploy relations to speed up overall execution
     postgresql_config = {
-        "plugin_pg_trgm_enable": True,
-        "plugin_unaccent_enable": True,
+        "plugin_pg_trgm_enable": "True",
+        "plugin_unaccent_enable": "True",
```
jdkandersson, line 65: To help reduce spelling mistakes you can use `str(True)` that way you get the Python interpreter checking that `True` is spelled correctly rather than having to visually inspect it

src/database_observer.py
jdkandersson: @gregory-schiano can also add his thoughts

Generally speaking, properties should be quite simple transformations or ways of making something private. This looks like a complex function with the possibility of errors being raised (e.g., `self.database.relations[0].id`, what if there are no relations?). Perhaps we should also be a bit more defensive on checking for those kind of errors?

tests/integration/conftest.py
arturo-seijas: Good point

docs/reference/integrations.md
```diff
@@ -8,7 +8,8 @@ _Supported charms_: [postgresql-k8s](https://charmhub.io/postgresql-k8s), [postg
 Database integration is a required relation for the indico charm to supply structured data
 storage for Indico.
 
-Example db integrate command: `juju integrate indico postgresql-k8s:db`
+Example db relate command: `juju relate juju relate indico postgresql-k8s`
```
merkata, line 11: It has `juju relate` twice.

pyproject.toml
```diff
@@ -16,7 +16,7 @@ line-length = 99
 target-version = ["py38"]
 
 [tool.coverage.report]
-fail_under = 95
+fail_under = 96
```
merkata, line 19: :clap:

src/database_observer.py
```diff
@@ -0,0 +1,72 @@
...
+            charm: The parent charm to attach the observer to.
+        """
+        super().__init__(charm, "database-observer")
+        self._charm = charm
+        # SUPERUSER is required for the schema migrations
+        self.database = DatabaseRequires(
+            self._charm,
+            relation_name=self._RELATION_NAME,
+            database_name=self._charm.app.name,
+            extra_user_roles="SUPERUSER",
```
merkata, line 41: That could be moved as a constant like _SCHEMA_MIGRATION_USER_NAME or something (just a thought, no need to change anything).

merkata (dismissed):
LGreatTM

docs/reference/integrations.md
arturo-seijas: Good catch

src/database_observer.py
gregory-schiano: The business logic is not very complex here as it's just fetching the info needed to build the URI string, so I think that's fine to have it as a property
BUT a property shouldn't raise any exception, therefore I agree with David that we should be very defensive to avoid any KeyError.
So adding few checks on relation_id and relation_data should be enough

merkata (dismissed):
:clap: well done

docs/reference/integrations.md
```diff
@@ -8,7 +8,8 @@ _Supported charms_: [postgresql-k8s](https://charmhub.io/postgresql-k8s), [postg
 Database integration is a required relation for the indico charm to supply structured data
 storage for Indico.
 
-Example db integrate command: `juju integrate indico postgresql-k8s:db`
+Example db relate command: `juju relate indico postgresql-k8s`
```
mthaddon, line 11: I don't think we should change this, this is the juju 3 version of the command

docs/tutorial/getting-started.md
```diff
@@ -64,7 +64,7 @@ indico/0*                 waiting   idle   10.1.74.70             Waiting for da
 Provide integration between Indico and PostgreSQL:
 
 ```bash
-juju integrate indico postgresql-k8s:db
+juju relate indico postgresql-k8s
```
mthaddon, line 67: Same as above

mthaddon (commented):
Two minor doc comments

mthaddon (dismissed):
LGTM, thx!

amandahla (dismissed):
LGTM, just left a commeng regarding how we are dealing with a common thing between charm in different ways regarding raising Exceptions or not.
https://github.com/canonical/synapse-operator/blob/main/src/database_observer.py#L121

merkata (approved):

mthaddon (approved):

PR #338, opened 2023-10-27
Retrieve SMTP config from SMTP integrator

arturo-seijas:
Applicable spec: N/A

### Overview

Retrieve SMTP config from SMTP integrator

### Rationale

Leverage the new SMTP integrator charm

### Module Changes

* New smtp observer module
* Changes in state
* Changes in charm

### Library Changes

* Added smtp lib

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ line-length = 99
 target-version = ["py38"]
 
 [tool.coverage.report]
-fail_under = 96
+fail_under = 97
 show_missing = true
 
 # Linting tools configuration
--- a/src/charm.py
+++ b/src/charm.py
@@ -72,7 +72,9 @@ class IndicoOperatorCharm(CharmBase):
         self.database = DatabaseObserver(self)
         self.smtp = SmtpObserver(self)
         try:
-            self.state = State.from_charm(self)
+            self.state = State.from_charm(
+                self, smtp_relation_data=self.smtp.smtp.get_relation_data()
+            )
         except CharmConfigInvalidError as exc:
             self.unit.status = ops.BlockedStatus(exc.msg)
             return
@@ -565,12 +567,12 @@ class IndicoOperatorCharm(CharmBase):
             },
         }
 
-        if self.smtp.smtp_config:
-            env_config["SMTP_LOGIN"] = self.smtp.smtp_config.login
-            env_config["SMTP_PASSWORD"] = self.smtp.smtp_config.password
-            env_config["SMTP_PORT"] = self.smtp.smtp_config.port
-            env_config["SMTP_SERVER"] = self.smtp.smtp_config.host
-            env_config["SMTP_USE_TLS"] = self.smtp.smtp_config.use_tls
+        if self.state.smtp_config:
+            env_config["SMTP_LOGIN"] = self.state.smtp_config.login
+            env_config["SMTP_PASSWORD"] = self.state.smtp_config.password
+            env_config["SMTP_PORT"] = self.state.smtp_config.port
+            env_config["SMTP_SERVER"] = self.state.smtp_config.host
+            env_config["SMTP_USE_TLS"] = self.state.smtp_config.use_tls
 
         # Required for monitoring Celery
         celery_config = {"worker_send_task_events": True, "task_send_sent_event": True}
--- a/src/charm.py
+++ b/src/charm.py
@@ -32,6 +32,7 @@ from ops.model import (
 from ops.pebble import ExecError
 
 from database_observer import DatabaseObserver
+from smtp_observer import SmtpObserver
 from state import CharmConfigInvalidError, ProxyConfig, State
 
 logger = logging.getLogger(__name__)
@@ -69,6 +70,7 @@ class IndicoOperatorCharm(CharmBase):
         """
         super().__init__(*args)
         self.database = DatabaseObserver(self)
+        self.smtp = SmtpObserver(self)
         try:
             self.state = State.from_charm(self)
         except CharmConfigInvalidError as exc:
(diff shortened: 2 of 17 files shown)


src/charm.py
```diff
@@ -558,16 +560,18 @@ def _get_indico_env_config(self, container: Container) -> Dict:
...
                 "default": "fs:/srv/indico/archive",
             },
         }
 
+        if self.smtp.smtp_config:
+            env_config["SMTP_LOGIN"] = self.smtp.smtp_config.login
+            env_config["SMTP_PASSWORD"] = self.smtp.smtp_config.password
+            env_config["SMTP_PORT"] = self.smtp.smtp_config.port
+            env_config["SMTP_SERVER"] = self.smtp.smtp_config.host
+            env_config["SMTP_USE_TLS"] = self.smtp.smtp_config.use_tls
```
gregory-schiano, line 573: The parameters are in the state but not filled in during init, this should use the state and not the observer

src/smtp_observer.py
```diff
@@ -0,0 +1,51 @@
...
+            event: the relation event.
+        """
+        self.smtp_config = SmtpConfig(
+            host=event.host,
+            port=event.port,
+            login=event.user,
+            password=event.password,
+            use_tls=event.transport_security is not TransportSecurity.NONE,
+        )
+        self._charm.on.config_changed.emit()
```
gregory-schiano, line 51: Even if the approach is interesting, I'd rather put the code of `_on_config_changed` in a method and call it from there instead of emitting the event

src/charm.py
arturo-seijas: I can't fill those during `init`. The relation might not exist

src/smtp_observer.py
arturo-seijas: That approach would require a huge refactor, that's why I've decided to emit the event instead, and keep refactoring as we start supporting other integrators

gregory-schiano: Why a huge refactor ? The content on `_on_config_changed` can be put in a dedicated method called from both sides.
From the observer you have access to the charm so you can call `self._charm.my_method()`

src/charm.py
gregory-schiano: Well if the relation doesn't exist during init, it won't exist later, so that's not a problem.

src/smtp_observer.py
arturo-seijas: I have access to the charm as a `CharmBase` instance

src/charm.py
arturo-seijas: I don't follow what you mean

pyproject.toml
```diff
@@ -16,7 +16,7 @@ line-length = 99
 target-version = ["py38"]
 
 [tool.coverage.report]
-fail_under = 96
+fail_under = 97
```
amandahla, line 19: not-related/minor: Shouldnt be 95?

arturo-seijas: That's the minimum as per our contributing guide, but since we've achieved 97 I decided to make it the new hard limit

src/smtp_observer.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        super().__init__(charm, "smtp-observer")
+        self._charm = charm
+        self.smtp = SmtpRequires(
+            self._charm,
+            relation_name=self._RELATION_NAME,
+        )
+        self.smtp_config = None
+        self.framework.observe(
+            self.smtp.on.smtp_data_available, self._saml_relation_data_available
```
amandahla, line 35: saml?

arturo-seijas: Ups. Fixed, thanks

src/smtp_observer.py
amandahla: Agree that is kind of a huge refactor, it would need to create some abstraction like the PebbleService here
https://github.com/canonical/synapse-operator/blob/dca3d242e8edd30034b9a4a7ea6eb620e54a3a5e/src/database_observer.py#L61
Isnt there a story for refactoring Indico applying the charm architecture guide? It could be changed there.

src/charm.py
amandahla: I believe it should have something similar to the saml_config [here](https://github.com/canonical/synapse-operator/blob/main/src/charm.py#L54C49-L54C49)
the State.from_charm is missing the smtp_config parameter.

arturo-seijas: That would be an option, but I won't be able to leverage the data validation provided by the lib with that approach. I'm not sure it leads to simpler and/or more cohesive code either

gregory-schiano: Can't we add a method to the library that returns the config as a dict and uses the getter to trigger validation?

src/smtp_observer.py
arturo-seijas: Are you sure you are linking the appropriate example? I don't see how that bypasses the limitation pointed out above

src/charm.py
arturo-seijas: That's definitely possible, but we decided to include in the library a mechanism for validating the relation data on the requirer side and we are going to bypass it?
```
if self._is_relation_data_valid(event.relation):
    self.on.smtp_data_available.emit(event.relation, app=event.app, unit=event.unit)
```

yanksyoon: I've done stateful initialization of relation data, [reference](https://github.com/canonical/jenkins-k8s-operator/blob/f4a1eb34d6528fb4ac69145d3db0b5807e3386a4/src/state.py#L216-L217). It can become a little verbose in the state.py.

arturo-seijas: Changed

amandahla:
One last question: Is SMTP required by Indico? Should be added to _are_relations_ready?

arturo-seijas:
> One last question: Is SMTP required by Indico? Should be added to _are_relations_ready?

Indico can work without an SMTP server

amandahla (dismissed):
LGTM

src/smtp_observer.py
yanksyoon: A passing thought, I think libraries emitting their own custom events like `smtp_data_available` should be considered antipattern. What about if a library exposes it's state?

yanksyoon (dismissed):
LGTM, I do think that emitting `config_changed` event should be refactored as soon as it can be :p

src/smtp_observer.py
arturo-seijas: Let's discuss this further, but not in this PR. Thanks for the feedback :)

gregory-schiano (approved):
LGTM

nrobinaubertin (approved):

amandahla (approved):
LGTM

Thanhphan1147 (approved):

PR #183, opened 2022-11-18
Downgrade charmcraft bases to ubuntu 20.04

arturo-seijas:
https://bugs.launchpad.net/juju/+bug/1996590

PR #361, opened 2024-02-15
Add run-on for jammy

Thanhphan1147:
### Checklist

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -11,3 +11,5 @@ bases:
     run-on:
     - name: "ubuntu"
       channel: "20.04"
+    - name: "ubuntu"
+      channel: "22.04"


mthaddon:
The reason we haven't done this so far, and I still don't think we want to do it is that if someone has deployed the charm on 20.04 and then we publish a version with 22.04 if they upgrade they'll still be on 20.04. This is related to https://bugs.launchpad.net/juju/+bug/1996590. See https://github.com/canonical/indico-operator/pull/183. If there's a way to remove the revision that has 22.04 to avoid future confusion that would be best - perhaps we can ask on discourse.charmhub.io?

Thanhphan1147:
I see, I'll close this PR and reach out to charmhub

PR #366, opened 2024-03-13
Deploy only one celery instance

arturo-seijas:
Applicable spec: <link> N/A

### Overview

Run celery only in one unit. To achieve that, the celery container is removed and it will be planned instead on a single unit along with the indico process

### Rationale

Celery does not provide exactly once guarantees, so multiple units can cause tasks to be executed by multiple workers, impacting user experience when those have a visible result.

### Juju Events Changes

* leader elected
* peer relation departed

### Module Changes

charm.py

### Library Changes

N/A

### Checklist


src/charm.py
```diff
@@ -457,9 +469,8 @@ def _get_indico_secret_key_from_relation(self) -> Optional[str]:
             secret_value = peer_relation.data[self.app].get("secret-key")
```
yanksyoon, line 469: Out of the scope of this PR but out of curiosity, may I ask what secret this is referring to? (Perhaps we can start naming secrets as well)

pyproject.toml
```diff
@@ -16,7 +16,7 @@ line-length = 99
 target-version = ["py38"]
 
 [tool.coverage.report]
-fail_under = 97
+fail_under = 96
```
yanksyoon, line 19: I'm guessing the drop is from the pebble layer changes?

yanksyoon (dismissed):
LGTM! Just questions out of curiousity!

pyproject.toml
arturo-seijas: Oops, I've reverted that change

yanksyoon (approved):

mthaddon (commented):
If I'm understanding this correctly we're relying on the `peer-relation-departed` event happening before `leader-elected`. If it doesn't, we'd remove the nominated unit but not nominate a new unit until another `leader-elected` event is triggered at some point in the future.

arturo-seijas:
> If I'm understanding this correctly we're relying on the `peer-relation-departed` event happening before `leader-elected`. If it doesn't, we'd remove the nominated unit but not nominate a new unit until another `leader-elected` event is triggered at some point in the future.

Exactly, there should be a leader at any point, so if the celery unit is leader, a leader elected event will be fired when it is removed (if not earlier)

mthaddon:
> Exactly, there should be a leader at any point, so if the celery unit is leader, a leader elected event will be fired when it is removed (if not earlier)

But if the `leader-elected` event is fired before `peer-relation-departed` we'd run the following:
```
if peer_relation and not peer_relation.data[self.app].get("celery-unit"):
```
Since "celery-unit" is set, it would do nothing. Then the `peer-relation-departed` event fires, and we unset this value, and we'd continue with an unset value until the next `leader-elected` event is fired, at some unknown point in the future.

arturo-seijas:
> > Exactly, there should be a leader at any point, so if the celery unit is leader, a leader elected event will be fired when it is removed (if not earlier)
> 
> But if the `leader-elected` event is fired before `peer-relation-departed` we'd run the following:
> 
> ```
> if peer_relation and not peer_relation.data[self.app].get("celery-unit"):
> ```
> 
> Since "celery-unit" is set, it would do nothing. Then the `peer-relation-departed` event fires, and we unset this value, and we'd continue with an unset value until the next `leader-elected` event is fired, at some unknown point in the future.

In that case, the change would occur in the peer relation departed handler. All the units receive that event, so if the celery unit is not the leader, the leader will set that value to itself

mthaddon:
> In that case, the change would occur in the peer relation departed handler. All the units receive that event, so if the celery unit is not the leader, the leader will set that value to itself

Sorry, you are right. Thanks for the out of band explanation, I'd missed that the `peer-relation-departed` event is handling things different if the "celery-unit" value is set to the current leader or not exactly to take care of this possibility.

tests/integration/test_charm.py
jdkandersson, line 1: Do you think we should have a test where the leader is changed? E.g., deploy 2 units, check that celery is working, kill the current leader and wait for the next leader and then check that celery has moved to the new leader unit?

arturo-seijas: It seems to me that use case is already covered by the unit tests

javierdelapuente (approved):

gregory-schiano (approved):

PR #447, opened 2024-08-29
feat: Updating Redis relations

alithethird:
Updating the charm according to latest [Redis PR](https://github.com/canonical/redis-k8s-operator/commit/63b8cfa88196a92f913ef05eba7c3fd8f6e8d1a6)
@gregory-schiano 
### Overview
Now we get the Url for redis instances from App databag instead of unit databag.

### Rationale
This fixes the issue when Redis is in HA there was a chance the Redis Url used in Indico charm would not have write access to the Redis db.

### Checklist

- [ ] The documentation for charmhub is updated


arturo-seijas (dismissed):

src/charm.py
```diff
@@ -380,6 +381,36 @@ def _get_nginx_pebble_config(self, _) -> ops.pebble.LayerDict:
...
+            Url for the redis-broker charm.
+        """
+        relation = self.model.get_relation(self.redis_broker.relation_name)
+        if relation:
+            relation_data = relation.data[self.model.relations["redis-broker"][0].app]
+            ip = relation_data["leader-host"]
+            unit_relation = self.redis_broker.relation_data
+            port = unit_relation["port"]
+            return f"redis://{ip}:{port}"
+        return ""
```
mthaddon, line 412: It seems like it might be worth refactoring this to a single function that you pass the relation name to?

src/charm.py
```diff
@@ -380,6 +381,36 @@ def _get_nginx_pebble_config(self, _) -> ops.pebble.LayerDict:
...
+        Returns:
+            Url for the redis-cache charm.
+        """
+        relation = self.model.get_relation(self.redis_cache.relation_name)
+        if relation:
+            relation_data = relation.data[self.model.relations["redis-cache"][0].app]
+            ip = relation_data["leader-host"]
+            unit_relation = self.redis_cache.relation_data
+            port = unit_relation["port"]
+            return f"redis://{ip}:{port}"
```
mthaddon, line 396: It seems we're getting the ip address from the application databag, but the port from the `relation_data` property of the redis_k8s library, which just grabs the "first" unit data (see https://github.com/canonical/redis-k8s-operator/blob/main/lib/charms/redis_k8s/v0/redis.py#L82-L92). That seems a bit risky, as we're not actually confirming the port from the same unit that happens to be the "leader-host" currently. In practice it's probably always going to use the same port for all units, but we're introducing a reliance on that, which we may want to avoid. Could be worth discussing this with others as well to determine if we're okay with that risk.

tests/unit/test_base.py
```diff
@@ -38,10 +38,16 @@ def set_up_all_relations(self):
...
+            unit_data={"hostname": "broker-host", "port": "1010"},
+            app_data={"leader-host": "broker-host"},
         )
         self.harness.add_relation(
-            "redis-cache", "redis-cache", unit_data={"hostname": "cache-host", "port": "1011"}
+            "redis-cache",
+            "redis-cache",
+            unit_data={"hostname": "cache-host", "port": "1011"},
+            app_data={"leader-host": "cache-host"},
         )
```
mthaddon, line 51: Could we see some test data when we have more than one unit, to see how that's handled?

src/charm.py
```diff
@@ -80,7 +80,8 @@ def __init__(self, *args):
         self.framework.observe(self.on.indico_pebble_ready, self._on_pebble_ready)
         self.framework.observe(self.on.indico_nginx_pebble_ready, self._on_pebble_ready)
         self.framework.observe(
-            self.on.refresh_external_resources_action, self._refresh_external_resources_action
+            self.on.refresh_external_resources_action,
+            self._refresh_external_resources_action,
```
mthaddon, line 84: Is this change needed? Seems unrelated to this PR.

gregory-schiano: Seems like an IDE line length parameter that differs

src/charm.py
gregory-schiano: There"s definitely code duplication here

src/charm.py
gregory-schiano: I think the risk of having different ports is limited. Also we need to do that due to a lack of support from the library, so this is already a workaround

src/charm.py
```diff
@@ -380,6 +381,36 @@ def _get_nginx_pebble_config(self, _) -> ops.pebble.LayerDict:
...
+
+        Returns:
+            Url for the redis-cache charm.
+        """
+        relation = self.model.get_relation(self.redis_cache.relation_name)
+        if relation:
+            relation_data = relation.data[self.model.relations["redis-cache"][0].app]
+            ip = relation_data["leader-host"]
+            unit_relation = self.redis_cache.relation_data
+            port = unit_relation["port"]
```
gregory-schiano, line 395: You should be a bit defensive here when accessing keys in a dict, either use `.get` with a default value, or catch the KeyError

src/charm.py
alithethird: Thank you! Turned it into 1 function with redis name as input.

Should I return "" when redis name is wrong or throw an exception?

```python
    def _get_redis_url(self, redis_name: str) -> str:
        """Get Url for redis charm.

        Args:
            redis_name (str): Name of the redis charm to connect to.

        Returns:
            Url for the redis charm.
        """
        if redis_name == "redis-broker":
            redis = self.redis_broker
        elif redis_name == "redis-cache":
            redis = self.redis_cache
        else:
            return ""

        relation = self.model.get_relation(redis.relation_name)
        if relation:
            relation_data = relation.data[self.model.relations[redis_name][0].app]
            ip = relation_data["leader-host"]
            unit_relation = redis.relation_data
            port = unit_relation["port"]
            return f"redis://{ip}:{port}"
        return ""
```

src/charm.py
alithethird: I used the tox formatting for that and made sure no other formetter touched it but this is not necessary i guess. Should I revert it since its unnecessary?

src/charm.py
alithethird: I think its pretty hard to get the port of the redis Master. There is no straight way of getting the port from library. Only way I can think of is getting all the unit databags and take the one that has the same ip as master and get its port but the ip of master doesn't show as a regular ip (ex.`10.1.123.12`) it shows as `redis-broker-0.redis-broker-endpoints.indico.svc.cluster.local` so I guess this would not work either

tests/unit/test_base.py
alithethird: That's a great idea thanks! I add some and previous tests are still passing. I also want to add a test that checks when new master unit adds. I will work on that if you think that is a good idea.

src/charm.py
alithethird: Thanks! Now catching the `KeyError`

src/charm.py
mthaddon: I think you should raise an exception, because it would definitely be unexpected

src/charm.py
```diff
@@ -380,6 +384,37 @@ def _get_nginx_pebble_config(self, _) -> ops.pebble.LayerDict:
...
+        else:
+            raise InvalidRedisNameError(f"Invalid Redis name: {redis_name}")
+
+        relation = self.model.get_relation(redis.relation_name)
+        if relation:
+            try:
+                relation_data = relation.data[self.model.relations[redis_name][0].app]
+                ip = relation_data["leader-host"]
+                unit_relation = redis.relation_data
+                port = unit_relation["port"]
```
mthaddon, line 412: I think you could collapse these two lines into: `port = redis.relation_data["port"]`

alithethird: Yeah you are right, thanks!

mthaddon (dismissed):
LGTM, thx

mthaddon (approved):
LGTM

amandahla (approved):
LGTM

